PR #2, opened 2023-02-04
feat: add WebSOC and Registrar API modules

ecxyzzy:


y-dejong (approved):
Looks good to me! Websoc and registrar tests passing

PR #10, opened 2023-02-27
docs: create contributors list

ecxyzzy:
Req'ing Raman and Mars's reviews here since I'm not familiar with the Ancient™ PeterPortal™ Lore™ and want to give proper credit where it's due. As I understand it API and Core split off at some point, so I'm not sure who would be considered project lead for and/or founder of API before that, if anyone at all.

Also just wondering whether Mars is cool with the list listing Mars as their name. (Wack sentence for sure.)

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -4,10 +4,11 @@ Contributors are listed below in alphabetical order of their GitHub username.
 
 ## Project Leads
 
-| Year      | Project Lead                                         |
-| --------- | ---------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


ramanxg:
I think we can include @chenaaron3, since he worked a lot on the original scraper. Other than that, mars was the founder of the project. Awesome to see this!

uci-mars (approved):
love this!

+1 to adding @chenaaron3

uci-mars:
and yes Mars is cool with the list listing Mars as his name

PR #12, opened 2023-03-02
feat: add Docusaurus boilerplate and initial relevant documentation (WIP)

ecxyzzy:


ecxyzzy:
Recreated this PR in #13 because its history got too complicated and would be a nightmare to review.

PR #18, opened 2023-03-08
feat: Query schema

bevm0:
Added Zod to parse the query

- [`schema.parse`](https://github.com/colinhacks/zod#parse) is used to validate the unknown query and statically type its output
- [`schema.safeParse`](https://github.com/colinhacks/zod#safeparse) can also be used parse the unknown but prevent throwing an error so post-processing can be done on the input if needed
- throws a ZodParseError if invalid, and the message can be configured per object property, refinement, etc.
- better TypeScript support and much more succinct than doing native runtime validation

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "prettier-plugin-packagejson": "^2.4.0",
     "prettier-plugin-prisma": "^4.8.0",
     "prettier-plugin-sql": "^0.12.1",
-    "turbo": "^1.7.0",
+    "turbo": "^1.8.3",
     "typescript": "^4.9.4"
   },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -56,7 +56,7 @@
     "prettier-plugin-packagejson": "^2.4.0",
     "prettier-plugin-prisma": "^4.8.0",
     "prettier-plugin-sql": "^0.12.1",
-    "turbo": "^1.7.0",
+    "turbo": "^1.8.3",
```
bevm0, line 59: for some reason, turbo wouldn't install for me until I bumped the version up

PR #20, opened 2023-03-09
refactor: simplify the websoc combine process

bevm0:
# Summary
Attempt to reduce the complexity of the combining function.
Proposal / demo of how it could look instead.
Needs testing and verification because I've never actually seen or interacted with the `WebsocResponse[]`

# Strategy
1. Use `map` and `flat` to generate a flattened array of enhanced sections: sections that have contextual metadata about the school, department, and course they're in
2. Use `reduce` to create an array of schools by iterating through enhanced sections and handling each appropriately.
- Attempt to find some "part" of the section in the accumulated array of schools
- If it doesn't exist, then add that "part" of the section to the corresponding property of the accumulated schools
- Otherwise, index into the found "part" and repeat

# Notes
- Doesn't replace the current implementation yet because idk how to test it
- This solution intends to "abuse" object references to `push` to the correct arrays, but I'm not sure how this actually plays out in reality -- i.e. where we should make deep copies or keep original references

# Why
- Easier to read and reason about for my smol brain
- No more dependence on the external `object-hash` library
- Better utilizes TypeScript null checking and guarding to prevent superfluous optional chaining and re-indexing of previously processed arrays

### DIFF
--- a/api/routes/websoc/websoc.service.ts
+++ b/api/routes/websoc/websoc.service.ts
@@ -58,7 +58,7 @@ const isolateSection = (data: EnhancedSection): EnhancedSection => {
  * @param responses The responses to combine.
  */
 export const combineResponses = (
-  responses: WebsocAPIResponse[]
+  ...responses: WebsocAPIResponse[]
 ): WebsocAPIResponse => {
   /**
    * sections are enhanced with context of parent structures and unique meetings
--- a/api/routes/websoc/websoc.service.ts
+++ b/api/routes/websoc/websoc.service.ts
@@ -1,4 +1,3 @@
-import hash from "object-hash";
 import type {
   WebsocAPIResponse,
   WebsocCourse,
@@ -27,7 +26,7 @@ export const combineResponses = (
   responses: WebsocAPIResponse[]
 ): WebsocAPIResponse => {
   /**
-   * all sections are enhanced with contextual info from parent structures
+   * sections are enhanced with context of parent structures and unique meetings
    */
   const allSections: EnhancedSection[] = responses
     .map((response) =>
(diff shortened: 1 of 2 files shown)


api/routes/websoc/websoc.service.ts
```diff
@@ -0,0 +1,125 @@
...
+    )
+    .flat();
+
+  /**
+   * for each section:
+   * if one of its parent structures hasn't been declared,
+   * append that structure appropriately
+   */
+  const schools = allSections.reduce((acc, section) => {
+    const foundSchool = acc.find(
```
bevm0, line 63: Start by accessing the accumulated array of schools. Every time a parent structure is found, index the previous calculated reference to easily find a sub-property of the accumulated array. Calling `push` on this sub property *should* mutate accumulated array in the way we want?

api/routes/websoc/websoc.service.ts
```diff
@@ -0,0 +1,125 @@
...
+          const meetingsHashSet: Record<string, WebsocSectionMeeting> = {};
+          for (const meeting of e.meetings) {
+            const meetingHash = hash([meeting.days, meeting.time]);
+            if (meetingHash in meetingsHashSet) {
+              meetingsHashSet[meetingHash].bldg.push(meeting.bldg[0]);
+            } else {
+              meetingsHashSet[meetingHash] = { ...meeting };
+            }
+            e.meetings = Object.values(meetingsHashSet);
+          }
```
bevm0, line 118: Is it possible to calculate this during one of the previous steps, or to remove the dependence on `object-hash`?

ecxyzzy (commented):
LGTM so far. You could test it by replacing the calls in the route with the calls in the helper file, then committing the modification and making requests to the staging endpoint.

api/routes/websoc/websoc.service.ts
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * preserves context for each section
+ */
+interface EnhancedSection {
+  school: WebsocSchool;
+  department: WebsocDepartment;
+  course: WebsocCourse;
+  section: WebsocSection;
+}
```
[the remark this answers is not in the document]
bevm0: jk, it can be kept around for the `isolate` helper function

bevm0:
Ok, it seems to generate the same response as the production website.
image

api/routes/websoc/websoc.service.ts
bevm0: It seems like it's doing the correct thing when I test the staging endpoint.

ecxyzzy (commented):
LGTM, already incorporated into #21 so I'll be closing this.

PR #21, opened 2023-03-10
feat: implement WebSoc scraping and RDS-backed endpoint

### LINKED ISSUE #11: Investigate feasibility of proactive WebSoc scraping
This was originally worked on in #1, but was deemed infeasible for a multitude of reasons.

- As of the time of posting, there are 143 distinct departments available. If each department needs to be scraped every 5 minutes, and there can be up to 5 distinct terms for which to scrape at any given time (Spring, three Summer Sessions, Fall), using highly parallelized Lambdas (like in the original implementation) would result in ridiculous cost overruns. A long-running EC2/Fargate instance might work, but WebSoc is slow, and may result in some portions of the data being scraped in the current run already going stale before the next run begins.
- Even disregarding the tremendous cost of maintaining a reasonably up-to-date cache of pertinent WebSoc data, we would have to essentially reimplement WebSoc's filtering in TypeScript and/or using a database. Given the structure of the data this is anywhere from a mild inconvenience to an insurmountable pain.

However, that's not to say this idea should be given up on entirely, so I'm creating this issue and putting it in the icebox if anyone wants to attempt to tackle it without bankrupting ICSSC.

ecxyzzy:
# Summary
Switch to using our Amazon RDS (managed relational database) instance for the WebSoc cache. This will allow us to cache all of WebSoc and serve arbitrary requests with it, rather than the limited set of queries we currently support.

Special thanks to @MinhxNguyen7 for helping with brainstorming and testing for this feature.

TODO:
- [x] Add full courses and cancelled courses filtering
- [x] Incorporate changes from #20 
- [x] Add logic for scraper
- [x] ~~Determine optimal solution for hosting scraper (EC2)~~
- [x] ~~Rewrite scraper~~
- [x] ~~Determine code deployment process for that solution~~
- [x] ~~Implement code deployment process~~
- [x] #26 

# Issues

Closes #11.

# Future Followup

The scraper can serve as a basis for getting enrollment data for that endpoint.


tools/websoc-api-next/index.ts
```diff
@@ -11,6 +12,7 @@ import {
   Term,
   TermData,
   WebsocAPIResponse,
+  WebsocSectionMeeting,
 } from "peterportal-api-next-types";
 
 /* region Constants */
```
bevm0, line 18:
```suggestion
//-----------------------------------------------------------------------------------
// START Constants
//-----------------------------------------------------------------------------------
```
My engineering verilog classes had lots of comments like this and they've kinda grown on me lol

MinhxNguyen7:
How much faster is that?

MinhxNguyen7:
How much faster is that?

api/routes/websoc/build.mjs
```diff
@@ -29,6 +29,7 @@ import { fileURLToPath } from "url";
         name: "copy",
         setup(build) {
           build.onEnd(async () => {
+            // camaro
             await Promise.all(
```
bevm0, line 33: Can this array be stored as a variable at the top of the file with a descriptive comment?

api/routes/websoc/build.mjs
```diff
@@ -50,6 +51,22 @@ import { fileURLToPath } from "url";
                 )
               )
             );
+            // prisma
+            await copyFile(
+              join(
+                cwd,
+                "../../../node_modules/.prisma/client/libquery_engine-rhel-openssl-1.0.x.so.node"
```
bevm0, line 58: If possible, I'd like to look into somehow using node resolution to resolve these paths, i.e. the same resolution that `yarn` uses to somehow find the prisma schema.

For now, a compromise might be to store these strings as descriptive constants, e.g.
```ts
/**
 * Output of `prisma generate`
 */
const generatedPrisma = '../../node_modules/.prisma/client`
```

api/routes/websoc/build.mjs
bevm0, line 1: I'm not a big fan of the "immediately invoked lambda function" pattern because there are too many parentheses lol. I think it's lot simpler to immediately parse and understand if it's a vanilla function declaration + invocation. e.g.
```ts
async function buildApp() {
  // esbuild this app
}

// just call the function when the script runs
buildApp()
```

api/routes/websoc/build.mjs
```diff
@@ -29,6 +29,7 @@ import { fileURLToPath } from "url";
         name: "copy",
```
bevm0, line 29: It's up to preference, but could we have separate plugins for dealing with prisma vs. dealing with external dependencies?
We might be able to extract the prisma plugin altogether into its own monorepo package.

api/routes/websoc/schema.ts
```diff
@@ -0,0 +1,128 @@
...
+    ge: z.enum(anyArray).or(z.enum(geCodes)).optional(),
+    department: z.string().optional(),
+    courseTitle: z.string().optional(),
+    courseNumber: z
+      .string()
+      .array()
+      .or(z.string())
+      .optional()
+      .transform(normalizeValue)
+      .optional(),
```
bevm0, line 67: I'm not sure if we'd like to continue discussing this, but this doesn't feel like idiomatic schema usage.

The reason the `optional` is appended onto the end is to force the output type to be possibly undefined; this is fundamentally incorrect because this type should be entirely dependent on the return type of `normalizeValue`. 

I imagine the scenario is something like this: `courseNumber` can be a string, string array, or undefined. After we normalize it, it'll be a sorted array or undefined still. In that case, `normalizeValue` should have a return type that's union-ed with undefined. `optional` is generally meant to parse the input **before** the transformation step. 

TLDR: Chaining a final `optional` incorrectly type-asserts the return type of `normalizeValue`, and we should adjust `normalizeValue` to return `T | undefined`, and then drop all trailing `optional` calls after a `transform`.

api/routes/websoc/index.ts
```diff
@@ -1,572 +1,151 @@
...
-        // Check if each result fulfilled or rejected; merge the data of each
-        // fulfilled response into the final response and remove it from the
-        // array of queries. If there are still responses remaining, wait for 1
-        // second and repeat the process.
+
+        let queries = normalizeQuery(parsedQuery);
+        let websocResponseData: WebsocAPIResponse = { schools: [] };
+        let retries = 0;
+
         for (;;) {
```
bevm0, line 103:
```suggestion
        while (queries.length && retries < 3) {
```
Can we extract the "failed to respond error" outside? e.g.
```ts
while (queries.length && retries < 3) {
  // try fetching queries
}

// now that we're done with all the attempts, check if an error condition has occurred
if (queries.length && retries >= 3) {
  createErrorResult(...)
}
```
This simplifies the logic in the loop body.

api/routes/websoc/index.ts
```diff
@@ -1,572 +1,151 @@
...
-        // Collect them into an iterable of ``Promises`` and fire them in parallel.
-        // Check if each result fulfilled or rejected; merge the data of each
-        // fulfilled response into the final response and remove it from the
-        // array of queries. If there are still responses remaining, wait for 1
-        // second and repeat the process.
+
+        let queries = normalizeQuery(parsedQuery);
+        let websocResponseData: WebsocAPIResponse = { schools: [] };
+        let retries = 0;
+
```
bevm0, line 102:
```suggestion
let failed: WebsocAPIOptions[] = [];
```
We can move the failed array outside as well.

api/routes/websoc/index.ts
bevm0, line 1: Overall, this file looks much cleaner than the very first implementation. Good work! :rocket: :sunglasses: :fire:

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+    waitlistFull: boolean;
+    overEnrolled: boolean;
+    cancelled: boolean;
+    data: object;
+  };
+};
+
+// The duration to sleep between scraping runs, or to wait before querying the
+// database for terms to scrape on demand if there were none in the last query.
+const SLEEP_DURATION = 5 * 60 * 1000; // 5 minutes
```
bevm0, line 109:
```suggestion
/**
 * The duration to sleep between scraping runs, or to wait before querying the
 * database for terms to scrape on demand if there were none in the last query.
 * default: 5 minutes
 */
const SLEEP_DURATION = 5 * 60 * 1000;
```

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+  course: WebsocCourse;
+  section: WebsocSection;
+};
+
+/**
+ * Returns the lexicographical ordering of two elements.
+ * @param a The left hand side of the comparison.
+ * @param b The right hand side of the comparison.
+ */
+const lexOrd = (a: string, b: string): number => (a === b ? 0 : a > b ? 1 : -1);
```
bevm0, line 29: This would be significantly better to me as `lexOrder`, but I'm fully aware that prettier will force this into 2 lines with those extra characters lol :sob:

I don't really like the 80 default default print width for prettier, I think 90-100 characters is perfectly fine and much more standard.

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+ * type guard that asserts that the settled promise was fulfilled
+ */
+export const fulfilled = <T>(
+  value: PromiseSettledResult<T>
+): value is PromiseFulfilledResult<T> => value.status === "fulfilled";
+
+/**
+ * type guard that asserts input is defined
+ */
+export const notNull = <T>(x: T): x is NonNullable<T> => x != null;
```
bevm0, line 52: This honestly might be worth extracting into a completely miscellaneous package of utils, e.g.
```ts
// packages/utils/src/index.ts

/**
 * type guard that asserts input is defined
 */
export const notNull = <T>(x: T): x is NonNullable<T> => x != null;


// api/routes/websoc/lib.ts
import { notNull } from '@packages/utils'
```
Same for the sleep function.

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+      case "SkipFullWaitlist":
+        AND.push({ sectionFull: false, waitlistFull: false });
+    }
+  }
+
+  switch (parsedQuery.cancelledCourses) {
+    case undefined:
+    case "Exclude":
+      AND.push({ cancelled: false });
+      break;
```
bevm0, line 306:
```suggestion
    case undefined:
       AND.push({ cancelled: false });
      break;
    case "Exclude":
      AND.push({ cancelled: false });
      break;
```
I get what it's supposed to do, but I think it's nice to have defined behavior for each case, even if they're duplicated.

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+      AND.push({ cancelled: true });
+  }
+
+  if (parsedQuery.sectionCodes) {
+    OR.push(
+      ...parsedQuery.sectionCodes.map((code) => ({
+        sectionCode: code.includes("-")
+          ? {
+              gte: parseInt(code.split("-")[0]),
+              lte: parseInt(code.split("-")[1]),
```
bevm0, line 317:
```suggestion
              gte: parseInt(code.split("-")[0], 10),
              lte: parseInt(code.split("-")[1]), 10,
```

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+      .filter(notNull);
+
+    return query.units
+      .map((units) => ({ ...baseQuery, units }))
+      .flatMap((copiedQuery) =>
+        keys.map((k) => ({
+          ...copiedQuery,
+          sectionCodes: (query.sectionCodes ?? [])
+            .slice(k * 5, (k + 1) * 5)
+            .join(","),
```
bevm0, line 389:
```suggestion
          sectionCodes: query.sectionCodes?
            .slice(k * 5, (k + 1) * 5)
            .join(",") || [],
```
I generally don't like doing an early `or` because I could negate short circuiting. e.g. If `query.SectionCodes` is undefined, then it can immediately short circuit to `[]`, whereas creating the array early will result in it being sliced and joined.

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+    schools.departments.forEach((department) => {
+      department.courses.forEach((course) =>
+        course.sections.sort(
+          (a, b) => parseInt(a.sectionCode) - parseInt(b.sectionCode)
+        )
+      );
+      department.courses.sort((a, b) => {
+        const numOrd =
+          parseInt(a.courseNumber.replace(/\D/g, "")) -
+          parseInt(b.courseNumber.replace(/\D/g, ""));
```
bevm0, line 417:
```suggestion
          parseInt(a.courseNumber.replace(/\D/g, ""), 10) -
          parseInt(b.courseNumber.replace(/\D/g, ""), 10);
```

api/routes/websoc/lib.ts
```diff
@@ -0,0 +1,425 @@
...
+ * numerical order of their number (with lexicographical tiebreaks),
+ * and sections are sorted in numerical order of their code.
+ * @param response The response to sort.
+ */
+export function sortResponse(response: WebsocAPIResponse): WebsocAPIResponse {
+  response.schools.forEach((schools) => {
+    schools.departments.forEach((department) => {
+      department.courses.forEach((course) =>
+        course.sections.sort(
+          (a, b) => parseInt(a.sectionCode) - parseInt(b.sectionCode)
```
bevm0, line 411:
```suggestion
          (a, b) => parseInt(a.sectionCode, 10) - parseInt(b.sectionCode, 10)
```

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+ */
+async function getTermsToScrape(date: Date) {
+  const quarterDates = Object.assign(
+    {},
+    ...(await Promise.all(
+      Array.from(Array(3).keys()).map((x) =>
+        getTermDateData((x + date.getFullYear() - 1).toString())
+      )
+    ))
+  );
```
bevm0, line 145:
```suggestion
  const termDateData = await Promise.all(
    [
      ...(Array(3).keys()).map((x) =>
        getTermDateData((x + date.getFullYear() - 1).toString()
      )
    ])
  const quarterDates = {...termDateData}
```
I'm not sure of the reasoning behind the original code, but I was wondering if it could be simplified to something like this?

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+      .filter((term) => Object.keys(quarterDates).includes(term))
+      .filter((term) => date <= quarterDates[term].finalsStart)
+      .map(
+        (term) =>
+          [term, { year: term.split(" ")[0], quarter: term.split(" ")[1] }] as [
+            string,
+            Term
+          ]
+      )
+  );
```
bevm0, line 158:
```suggestion
  type TermEntry = [string, Term]

  const terms = await getTerms()
  const termEntries = terms
    .map((term) => term.shortName)
    .filter((term) => Object.keys(quarterDates).includes(term))
    .filter((term) => date <= quarterDates[term].finalsStart)
    .map((term) => {
      const termEntry: TermEntry = [term, { year: term.split(" ")[0], quarter: term.split(" ")[1] }]
      return termEntry
    })
  return Object.fromEntries(termEntries);
```
This sequence is more verbose and uses additional variables, but I think it really aids with logical clarity. It's also easier to view type definitions of each step to see how data is transformed to produce the output.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+ * The scraping function.
+ * @param name The name of the term to scrape.
+ * @param term The parameters of the term to scrape.
+ */
+async function scrape(name: string, term: Term) {
+  logger.info(`Scraping term ${name}`);
+  // The timestamp for this scraping run.
+  const timestamp = new Date();
+  // All departments to scrape.
+  const deptCodes = (await getDepts())
```
bevm0, line 211:
```suggestion
 const depts = await getDepts();
 
  // All departments to scrape.
  const deptCodes = depts
```

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+      }
+    }
+    inputs = inputs.filter((_, i) => !fulfilledIndices.includes(i));
+    logger.info(
+      `${fulfilledIndices.length} calls succeeded, ${inputs.length} remain`
+    );
+    if (!inputs.length) break;
+    logger.info("Sleeping for 1 minute");
+    await sleep(60 * 1000);
+  }
```
bevm0, line 256: Can you employ a similar strategy to the `api/routes/websoc/index.ts` file with the `failed` array, using a `while` loop instead of a `for` loop + `break`?

I like the `while` loop in these scenarios because it's immediately clear what the break condition is.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+  const res: Record<string, ProcessedSection> = {};
+  logger.info("Processing all sections");
+  for (const [term, data] of Object.entries(results)) {
+    for (const response of Object.values(data.department)) {
+      for (const school of response.schools) {
+        for (const department of school.departments) {
+          for (const course of department.courses) {
+            for (const section of course.sections) {
+              const [year, quarter] = term.split(" ") as [string, Quarter];
+              const sectionCode = parseInt(section.sectionCode);
```
bevm0, line 266:
```suggestion
              const sectionCode = parseInt(section.sectionCode, 10);
```

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                    name,
+                  })),
+                  meetings: section.meetings.map((m) => ({
+                    year,
+                    quarter,
+                    sectionCode,
+                    timestamp,
+                    days: ["Su", "M", "Tu", "W", "Th", "F", "Sa"].filter((x) =>
+                      m.days.includes(x)
+                    ),
```
bevm0, line 283: Extract the days array somewhere outside

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                          endTimeString.split(":");
+                        endTime =
+                          parseInt(endTimeHour) * 60 + parseInt(endTimeMinute);
+                        if (endTimeMinute.includes("p")) {
+                          startTime += 12 * 60;
+                          endTime += 12 * 60;
+                        }
+                      }
+                      return { startTime, endTime };
+                    })(),
```
bevm0, line 308: Extract this into a defined function; and make sure `parseInt` is always using base 10

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                  quarter,
+                  sectionCode: parseInt(section.sectionCode),
+                  timestamp,
+                  geCategories: [],
+                  department: department.deptCode,
+                  courseNumber: course.courseNumber,
+                  courseNumeric: (() => {
+                    const n = parseInt(course.courseNumber.replace(/\D/g, ""));
+                    return isNaN(n) ? 0 : n;
+                  })(),
```
bevm0, line 322: Extract this into a named function, or just inline the logic

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+    await sleep(60 * 1000);
+  }
+  const res: Record<string, ProcessedSection> = {};
+  logger.info("Processing all sections");
+  for (const [term, data] of Object.entries(results)) {
+    for (const response of Object.values(data.department)) {
+      for (const school of response.schools) {
+        for (const department of school.departments) {
+          for (const course of department.courses) {
+            for (const section of course.sections) {
```
bevm0, line 264: Prefer `forEach` when iterating over arrays. I *feel* like this could be expressed more succinctly as an accumulator + map function. Idk for sure tho, and I might just take a look at it myself later.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+              if (res[`${term} ${section.sectionCode}`]) {
+                res[`${term} ${section.sectionCode}`].data.geCategories.push(
+                  geCategory as GE
+                );
+              }
+            }
+          }
+        }
```
bevm0, line 357: Yeah same thing here, prefer `forEach` when iterating over arrays.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+  });
+  logger.info(`Removed ${sectionsDeleted.count} sections`);
+  logger.info("Sleeping for 5 minutes");
+  await sleep(SLEEP_DURATION);
+}
+
+/**
+ * The entry point of the program.
+ */
+(async function main() {
```
bevm0, line 405: Could we invoke this function normally at the end of the file?

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+  try {
+    logger.info("websoc-scraper-v2 daemon starting");
+    let now = new Date();
+    let termsInScope = await getTermsToScrape(now);
+    for (const [name, term] of Object.entries(termsInScope))
+      await scrape(name, term);
```
bevm0, line 411:
```suggestion
    Promise.all([
      Object.entries(termsInScope).map(([name, term]) => scrape(name, term);
    ])
```
Faster in parallel.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+    for (;;) {
+      const curr = new Date();
+      // Check the available terms every day and scrape all of them once.
+      if (curr.getDate() !== now.getDate()) {
+        termsInScope = await getTermsToScrape(curr);
+        for (const [name, term] of Object.entries(termsInScope))
+          await scrape(name, term);
```
bevm0, line 418: Same `Promise.all` strat I mentioned above.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+      now = curr;
+      // Check the database for terms to scrape on demand.
+      const termsOnDemand = Object.fromEntries(
+        (
+          await prisma.websocTerm.findMany({
+            where: { timestamp: { gte: now } },
+            select: { year: true, quarter: true },
+          })
+        ).map((x) => [`${x.year} ${x.quarter}`, x])
+      );
```
bevm0, line 429:
```suggestion
      const websocTerms =  await prisma.websocTerm.findMany({
          where: { timestamp: { gte: now } },
          select: { year: true, quarter: true },
        })
      const termsOnDemandEntries = websocTerms.map((x) => [`${x.year} ${x.quarter}`, x]);      
      const termsOnDemand = Object.fromEntries(termsOnDemandEntries);
```
Process is clearer.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+      let scraped = false;
+      for (const [name, term] of Object.entries(termsOnDemand)) {
+        if (!(name in termsInScope)) {
+          logger.info(`Term ${name} requested but not in scope, skipping`);
+          continue;
+        }
+        scraped = true;
+        await scrape(name, term);
+      }
```
bevm0, line 438: `Promise.all`

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+ * The entry point of the program.
+ */
+(async function main() {
+  try {
+    logger.info("websoc-scraper-v2 daemon starting");
+    let now = new Date();
+    let termsInScope = await getTermsToScrape(now);
+    for (const [name, term] of Object.entries(termsInScope))
+      await scrape(name, term);
+    for (;;) {
```
bevm0, line 412:
```suggestion
    while (true) {
```
Idk, I just don't like `for (;;)` lol.

I think it could be worth looking into using pm2's auto-restart with delay as opposed to an infinite loop. We'd be able to manage the instance and adjust the delay, restart conditions, etc. more easily. Encapsulating the entire program as an infinite loop just doesn't feel like the best long-term solution.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                  sectionType:
+                    section.sectionType as (typeof sectionTypes)[number],
+                  units: section.units,
+                  maxCapacity: parseInt(section.maxCapacity),
+                  sectionFull:
+                    section.status === "FULL" || section.status === "Waitl",
+                  waitlistFull: section.status === "FULL",
+                  overEnrolled:
+                    parseInt(section.numCurrentlyEnrolled.totalEnrolled) >
+                    parseInt(section.maxCapacity),
```
bevm0, line 333:
```suggestion
                    parseInt(section.numCurrentlyEnrolled.totalEnrolled, 10) >
                    parseInt(section.maxCapacity, 10),
```

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                  courseNumber: course.courseNumber,
+                  courseNumeric: (() => {
+                    const n = parseInt(course.courseNumber.replace(/\D/g, ""));
+                    return isNaN(n) ? 0 : n;
+                  })(),
+                  courseTitle: course.courseTitle,
+                  sectionType:
+                    section.sectionType as (typeof sectionTypes)[number],
+                  units: section.units,
+                  maxCapacity: parseInt(section.maxCapacity),
```
bevm0, line 327:
```suggestion
                  maxCapacity: parseInt(section.maxCapacity, 10),
```

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                        }
+                      }
+                      return { startTime, endTime };
+                    })(),
+                  })),
+                },
+                data: {
+                  year,
+                  quarter,
+                  sectionCode: parseInt(section.sectionCode),
```
bevm0, line 314:
```suggestion
                  sectionCode: parseInt(section.sectionCode, 10),
```

bevm0 (changes requested):
Overall, I think everything looks pretty solid. 👍 

I just wanted to discuss some of the comments I brought up before merging everything in.

api/routes/websoc/build.mjs
MinhxNguyen7: Yeah, I'd go as far as to argue that it's bad style because the syntax is not immediately clear.

tools/websoc-scraper-v2/index.ts
MinhxNguyen7: for(;;) is cursed

tools/websoc-scraper-v2/index.ts
MinhxNguyen7: We might get rate-limited if we do that. This ensures that we finish our request for a department before we ask for another one. Speed is not our friend when scraping WebSOC.

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+    for (const [i, res] of Object.entries(settledResults)) {
+      if (res.status === "fulfilled") {
+        const idx = parseInt(i);
+        const input = inputs[idx];
+        const term = `${input[0].year} ${input[0].quarter}`;
+        if (input[1].department) {
+          results[term].department[input[1].department] = res.value;
+        } else if (input[1].ge) {
+          results[term].ge[input[1].ge] = res.value;
+        }
```
MinhxNguyen7, line 245: I would have an `else` here and output a logger warning. I know that it _shouldn't_ be possible to have an input that has neither department nor ge, and that this would skip it, so it doesn't really matter, but WebdocAPIOptions does technically allow for it if there's an instructor search. I prefer to be explicit, but it's up to you.

MinhxNguyen7 (commented):
I don't have much to say other than style because I'm not that familiar with all this. Consider it an outsider's/newbie's perspective on your code. I only looked closely at the scraping code, and @bevm0 got most of what I was thinking. Otherwise, LGTM.

tsconfig.json
```diff
@@ -20,6 +20,7 @@
     "noUnusedParameters": false,
     "outDir": "dist",
     "preserveWatchOutput": true,
+    "resolveJsonModule": true,
```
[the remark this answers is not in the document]
ecxyzzy: I believe I may have added this while testing the scraper with JSON output. Is it bad practice to include this in a production-environment tsconfig?

tools/websoc-scraper-v2/index.ts
ecxyzzy: Minh is correct—we _don't_ want all scraping tasks to run in parallel because that will almost certainly get us ratelimited. Writing it this way ensures that each scraping task finishes its 5-minute wait before we start the next one.

api/routes/websoc/schema.ts
```diff
@@ -0,0 +1,128 @@
...
+function normalizeValue(value: NormalizeInput): string[] {
+  const unique = new Set(
+    Array.isArray(value)
+      ? value.flatMap((x) => x.split(","))
+      : value?.split(",")
+  );
+  return [...unique].sort();
+}
+
+const days = ["Su", "M", "Tu", "W", "Th", "F", "Sa"];
```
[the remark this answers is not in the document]
ecxyzzy: That's fine in a vacuum but WebSoc only recognizes these specific identifiers (among others that we don't intend to support) for the days of the week, so I think it's fine to hardcode them here.

api/routes/websoc/schema.ts
```diff
@@ -0,0 +1,128 @@
...
+  .object({
+    year: z
+      .string({ required_error: 'Parameter "year" not provided' })
+      .length(4, { message: "Invalid year provided" }),
+    quarter: z.enum(quarters, {
+      required_error: 'Parameter "quarter" not provided',
+      invalid_type_error: "Invalid quarter provided",
+    }),
+    ge: z.enum(anyArray).or(z.enum(geCodes)).optional(),
+    department: z.string().optional(),
```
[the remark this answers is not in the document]
ecxyzzy: That would require us to cache the list of available departments somewhere, or query WebSoc every time we got a request. I don't think it's necessary, since invalid department codes will just result in the `websoc-api-next` module returning `{ schools: [] }`.

db/prisma/schema.prisma
```diff
@@ -34,3 +34,84 @@ model GradesSection {
...
+}
+
+model WebsocSectionMeeting {
+  id          Int           @id @default(autoincrement())
+  year        String
+  quarter     Quarter
+  sectionCode Int
+  timestamp   DateTime
+  section     WebsocSection @relation(fields: [year, quarter, sectionCode, timestamp], references: [year, quarter, sectionCode, timestamp])
+  days        Json
```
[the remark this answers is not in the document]
ecxyzzy: MySQL/MariaDB does not support native scalar arrays. IMO it's unnecessary to build another many-to-many relation just for storing what day(s) a section meets on, since Prisma supports filtering on an array value for `Json` fields.

db/prisma/schema.prisma
```diff
@@ -34,3 +34,84 @@ model GradesSection {
...
+  startTime   Int
+  endTime     Int
+  buildings   Json
+}
+
+model WebsocTerm {
+  year      String
+  quarter   Quarter
+  timestamp DateTime
```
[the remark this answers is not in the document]
ecxyzzy: I believe the `timestamp` field is relatively standard to denote when the row was last updated. However, I can see how its meaning could be ambiguous here. Do you think `lastUpdated` would make more sense for this and `WebsocSection`?

ecxyzzy (commented):
I appreciate the feedback, especially the outsider perspective. One of the goals of Projects is to make the entry barrier for outside contributors as low as possible, and reviews in that view definitely help to work towards that goal.

api/routes/websoc/schema.ts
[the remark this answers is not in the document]
bevm0: Oh ok, I thought these were just arbitrary. Sounds good!

tsconfig.json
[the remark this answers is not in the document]
bevm0: No it's perfectly fine. I just like having a minimal tsconfig; I do that as a mini-game to learn more about what each option does and when I need it lol

tools/websoc-scraper-v2/index.ts
ecxyzzy: ESLint doesn't like constants in while loop conditions. That alone is probably a sign to refactor this into a script that runs only once and have pm2 manage the restart conditions.

ecxyzzy (commented):
Thanks GitHub for not auto-displaying large diffs and making me think that the files didn't have comments. Real helpful of you.

Everyone's comments should now be addressed.

api/routes/websoc/build.mjs
bevm0: It *is* a documented pattern https://developer.mozilla.org/en-US/docs/Glossary/IIFE.
So it's mostly up to preference/scenario dictating its use case.

db/prisma/schema.prisma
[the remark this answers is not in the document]
MinhxNguyen7: `lastUpdated` would make sense. Since `timestamp` is standard, I think it's fine to leave it like that well. It's up to you.

api/routes/websoc/build.mjs
MinhxNguyen7: Right, but none of the use cases (compared to definition + invocation) apply here. I suppose since it's documented, it's not "bad syntax" _per se_

api/routes/websoc/schema.ts
```diff
@@ -0,0 +1,127 @@
...
+      .or(z.string())
+      .optional()
+      .transform(flattenStringsAndSplit),
+    sectionCodes: z
+      .string()
+      .array()
+      .or(z.string())
+      .optional()
+      .transform(flattenStringsAndSplit)
+      .optional(),
```
bevm0, line 74:
```suggestion
      .transform(flattenStringsAndSplit),
```
`flattenStringsAndSplit` says it can return undefined, so I think this is redundant. Same for any other `optional` that are chained after `transform`.

api/routes/websoc/index.ts
```diff
@@ -1,572 +1,151 @@
...
- * Returns the lexicographical ordering of two elements.
- * @param a The left hand side of the comparison.
- * @param b The right hand side of the comparison.
- */
-const lexOrd = (a: string, b: string): number => (a === b ? 0 : a > b ? 1 : -1);
+export const rawHandler: RawHandler = async (request) => {
+  const { method, path, query, requestId } = request.getParams();
+  logger.info(`${method} ${path} ${JSON.stringify(query)}`);
+  switch (method) {
+    case "HEAD":
```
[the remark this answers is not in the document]
bevm0: Oh I didn't know that lol, ok sounds good

api/routes/websoc/index.ts
```diff
@@ -1,572 +1,151 @@
...
-          logger.info(``);
-          await sleep(1000);
+          // 3 attempts + (1 + 2 + 4) seconds ~= Lambda timeout (15 seconds)
+          if (retries >= 2)
+            return createErrorResult(
+              500,
+              "WebSoc failed to respond too many times. Please try again later.",
+              requestId
+            );
+          await sleep(1000 * 2 ** retries++);
```
bevm0, line 135: I'm confident the `createErrorResult` can be moved outside of the loop body, since the loop won't exceed that many retries before terminating. The `if (!queries.length) break` bothers me because it seems like we're defining a stop condition for the for loop that will never become true. A compromise might be to apply a condition to the sleep, e.g.
```ts
if (queries.length) await sleep(1000 * 2 ** retries++)
```

Now we have a for loop that has two valid conditions that can definitely be triggered (also this supports the reasoning that we want to move the error result outside), and we handle any other cases appropriately.

api/routes/websoc/lib.ts
bevm0: My bad, I forgot to respond to this. Resolved on Discord...

We can move internal packages to a root level `libs` directory, e.g. database, ddb, utils, etc.
Public packages can stay in the `packages` directory, e.g. peterportal-types

And if desired, we can configure scoping on internal packages to make them easier to find.
This entails:
1. Assigning their name in the package.json, e.g. `"name": "@libs/db"`
2. Configuring their path alias in the project root's tsconfig, e.g. compilerOptions -> paths.
3. Configuring other related settings in package.jsons? Idk how it works because I usually just define the path aliases manually for any build tools I use, instead of letting the monorepo handle it https://turbo.build/repo/docs/handbook/sharing-code/internal-packages#4-fix-main-and-types

bevm0 (changes requested):
Ok, this is a lot of content, but I left comments regarding code cleanup and project considerations. I'm not too aware of the actual state of the project at this point, so let me know if I can help out in testing or something. :rocket:

api/routes/websoc/index.ts
```diff
@@ -1,572 +1,159 @@
...
+          where: { year: parsedQuery.year, quarter: parsedQuery.quarter },
+        });
 
-/**
- * Returns the lexicographical ordering of two elements.
- * @param a The left hand side of the comparison.
- * @param b The right hand side of the comparison.
- */
-const lexOrd = (a: string, b: string): number => (a === b ? 0 : a > b ? 1 : -1);
+        if (parsedQuery.cache && termExists) {
```
bevm0, line 44: I changed this to "if cache enabled and term exists, find it in the database". I think that's the original intent?

api/routes/websoc/index.ts
```diff
@@ -1,572 +1,159 @@
...
+          });
+
+          successes = responses.filter(fulfilled);
+          websocResponseData = successes.reduce(
+            (acc, curr) => combineResponses(acc, curr.value),
+            websocResponseData
+          );
+
+          queries = failed;
+          if (queries.length) await sleep(1000 * 2 ** retries++);
```
bevm0, line 134: Now it *will* go back to the top after every loop, and hit the defined stop conditions. But a specific behavior per loop is to sleep if there are more upcoming queries, so I handled that accordingly.

api/routes/websoc/schema.ts
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Input to a transform function.
+ */
+type TransformInput = string | string[] | undefined;
+
+/**
+ * Output of a transform function.
+ */
+type TransformOutput = string[] | undefined;
```
bevm0, line 20: So I can squeeze the function defs on one line

api/routes/websoc/schema.ts
```diff
@@ -0,0 +1,126 @@
...
+ * @param value String of combined days of the week (e.g. ``MWF``) or array of such strings.
+ */
+function flattenDayStringsAndSplit(value: TransformInput): TransformOutput {
+  if (!value) return undefined;
+  const unique = new Set(
+    Array.isArray(value)
+      ? value.flatMap((x) => days.filter((y) => y.includes(x)))
+      : days.filter((x) => value.includes(x))
+  );
+  return [...unique];
```
bevm0, line 47: We don't actually need to sort the days, it's kind of a meme to do that; just turn it into an array

tools/websoc-scraper-v2/index.ts
```diff
@@ -0,0 +1,453 @@
...
+                  maxCapacity: parseInt(section.maxCapacity, 10),
+                  sectionFull:
+                    section.status === "FULL" || section.status === "Waitl",
+                  waitlistFull: section.status === "FULL",
+                  overEnrolled:
+                    parseInt(section.numCurrentlyEnrolled.totalEnrolled, 10) >
+                    parseInt(section.maxCapacity, 10),
+                  cancelled: section.sectionComment.includes(
+                    "***  CANCELLED  ***"
+                  ),
```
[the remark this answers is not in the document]
bevm0: uci moment

bevm0 (approved):
Ok I think it's good

PR #23, opened 2023-03-11
style: made code prettier

bevm0:
- Moved related functions closer to each other
- Made documentation terser
- I like function instead of const syntax lol

### DIFF
--- a/api/routes/websoc/index.ts
+++ b/api/routes/websoc/index.ts
@@ -1,15 +1,10 @@
+import type { LambdaHandler, RawHandler } from "api-core";
 import {
-  type IRequest,
   createErrorResult,
   createLambdaHandler,
   createOKResult,
   logger,
 } from "api-core";
-import type {
-  APIGatewayProxyEvent,
-  APIGatewayProxyResult,
-  Context,
-} from "aws-lambda";
 import { PrismaClient } from "db";
 import type { WebsocAPIResponse } from "peterportal-api-next-types";
 import { callWebSocAPI } from "websoc-api-next";
@@ -33,9 +28,7 @@ const fulfilled = <T>(
 
 const prisma = new PrismaClient();
 
-export const rawHandler = async (
-  request: IRequest
-): Promise<APIGatewayProxyResult> => {
+export const rawHandler: RawHandler = async (request) => {
   const { method, path, query, requestId } = request.getParams();
   logger.info(`${method} ${path} ${JSON.stringify(query)}`);
   switch (method) {
@@ -88,11 +81,6 @@ export const rawHandler = async (
       return createErrorResult(400, `Cannot ${method} ${path}`, requestId);
   }
 };
-
-type LambdaHandler = (
-  event: APIGatewayProxyEvent,
-  context: Context
-) => Promise<APIGatewayProxyResult>;
 
 export const lambdaHandler: LambdaHandler = async (event, context) =>
   createLambdaHandler(rawHandler)(event, context);
--- a/api/routes/websoc/index.ts
+++ b/api/routes/websoc/index.ts
@@ -59,11 +59,9 @@ export const rawHandler = async (
         }
 
         const queries = normalizeQuery(parsedQuery);
-
         const responses = await Promise.allSettled(
           queries.map((options) => callWebSocAPI(parsedQuery, options))
         );
-
         responses.forEach((response, i) => {
           const queryString = JSON.stringify(queries[i]);
           if (response.status === "fulfilled") {
@@ -74,12 +72,10 @@ export const rawHandler = async (
         });
 
         const successes = responses.filter(fulfilled);
-
         const websocResponseData = successes.reduce(
           (acc, curr) => combineResponses(acc, curr.value),
           { schools: [] } as WebsocAPIResponse
         );
-
         return createOKResult(sortResponse(websocResponseData), requestId);
       } catch (e) {
         if (e instanceof ZodError) {
@@ -93,8 +89,10 @@ export const rawHandler = async (
   }
 };
 
-export const lambdaHandler = async (
+type LambdaHandler = (
   event: APIGatewayProxyEvent,
   context: Context
-): Promise<APIGatewayProxyResult> =>
+) => Promise<APIGatewayProxyResult>;
+
+export const lambdaHandler: LambdaHandler = async (event, context) =>
   createLambdaHandler(rawHandler)(event, context);
--- a/api/routes/websoc/index.ts
+++ b/api/routes/websoc/index.ts
@@ -11,18 +11,25 @@ import type {
   Context,
 } from "aws-lambda";
 import { PrismaClient } from "db";
-import type { Term, WebsocAPIResponse } from "peterportal-api-next-types";
-import { type WebsocAPIOptions, callWebSocAPI } from "websoc-api-next";
-import type { ZodError } from "zod";
+import type { WebsocAPIResponse } from "peterportal-api-next-types";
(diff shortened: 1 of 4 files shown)


api/routes/websoc/schema.ts
```diff
@@ -61,16 +62,16 @@ export const QuerySchema = z
       .string()
       .array()
       .or(z.string())
-      .transform(normalizeValue)
-      .optional(),
```
bevm0, line 65: `transform` is one of the last things you call on the schema; `optional` has to be used before that point because Zod has to verify that it's valid *before* doing the transform

api/routes/websoc/schema.ts
```diff
@@ -10,41 +10,42 @@ import {
...
-  if (!val) return undefined;
-  const days = ["Su", "M", "Tu", "W", "Th", "F", "Sa"];
+function normalizeValue(value: NormalizeInput): string[] {
+  const unique = new Set(
+    Array.isArray(value)
+      ? value.flatMap((x) => x.split(","))
+      : value?.split(",")
+  );
+  return [...unique].sort();
+}
```
bevm0, line 28: I think this sequence reads more cleanly

api/routes/websoc/schema.ts
```diff
@@ -10,41 +10,42 @@ import {
...
-  Array.from(
-    new Set(
-      typeof val === "undefined"
-        ? [""]
-        : typeof val === "string"
-        ? val.split(",")
-        : val.flatMap((x) => x.split(","))
-    )
-  ).sort();
+type NormalizeInput = string | string[] | undefined;
```
bevm0, line 15: Utility type lol, now prettier fits some function signatures on one line

api/routes/websoc/lib.ts
```diff
@@ -72,39 +60,122 @@ const isolateSection = (data: EnhancedSection): EnhancedSection => {
...
+
+/**
+ * Converts a 12-hour time string into number of minutes since midnight.
+ * @param time The time string to parse.
+ */
+function minutesSinceMidnight(time: string): number {
+  const [hour, minute] = time.split(":");
+  return (
+    parseInt(hour, 10) * 60 +
+    parseInt(minute, 10) +
```
bevm0, line 146: Depending on your lint config, eslint will complain if you don't specify the base for parseInt

api/routes/websoc/index.ts
```diff
@@ -32,87 +39,60 @@ export const rawHandler = async (
...
 
-export const lambdaHandler = async (
+type LambdaHandler = (
   event: APIGatewayProxyEvent,
   context: Context
-): Promise<APIGatewayProxyResult> =>
+) => Promise<APIGatewayProxyResult>;
+
+export const lambdaHandler: LambdaHandler = async (event, context) =>
   createLambdaHandler(rawHandler)(event, context);
```
bevm0, line 98: If all lambda handlers have this signature, or a really similar one, then we should extract this to `api-core` or something (i.e. as a generic if necessary) and just type the function itself instead of its signature everytime

api/routes/websoc/lib.ts
```diff
@@ -72,39 +60,120 @@ const isolateSection = (data: EnhancedSection): EnhancedSection => {
...
       geCategories: {
         array_contains: parsedQuery.ge,
       },
     });
   }
+
   if (parsedQuery.department) {
-    AND.push({
-      department: parsedQuery.department,
-    });
```
bevm0, line 106: Sometimes it's more legible to squash lines of the prisma query into a single line. I did this wherever I felt like this applied

api/routes/websoc/schema.ts
```diff
@@ -81,9 +82,12 @@ export const QuerySchema = z
...
       .or(z.string())
-      .transform(normalizeValue)
-      .optional(),
-    cache: z.string().optional(),
+      .optional()
+      .transform(normalizeValue),
+    cache: z
+      .string()
+      .optional()
+      .transform((x) => !x || x !== "false"),
```
bevm0, line 90: I'm gonna convert this to a boolean, but I'm not sure if this will have tangible effects outside of this lambda handler

api/routes/websoc/schema.ts
ecxyzzy: The main issue (in my experience) with not having `optional` as the last method on the schema field is that the resulting field is not typed as optional. I ran into this pitfall while working on constructing the Prisma query, where the optional query params wouldn't error without optional chaining even though they were definitely intended to be optional.

api/routes/websoc/index.ts
ecxyzzy: `LambdaHandler` is already in `api-core` but isn't exported, feel free to do that though.

ecxyzzy (changes requested):
See comments, everything else looks good though.

Also RE: expression vs declaration syntax—while I agree that expression syntax looks better, IIRC they can't be typed, which would not be ideal for e.g. the Lambda handler. I also don't think there's an eslint rule that would warn on declaration syntax iff it's not otherwise typed, so I'm not really sure how best to proceed.

bevm0:
> See comments, everything else looks good though.
> 
> Also RE: expression vs declaration syntax—while I agree that expression syntax looks better, IIRC they can't be typed, which would not be ideal for e.g. the Lambda handler. I also don't think there's an eslint rule that would warn on declaration syntax iff it's not otherwise typed, so I'm not really sure how best to proceed.

In general, I defer to the `function` syntax because it's generally easier to reason about and parse, e.g. the following reads very cleanly.
```ts
function print(str: string): void { ... }
```

I only use the `const` syntax for several cases (not exhaustive):
1. In frontend web dev, event handlers can actually invoke `this` to refer to the DOM element that the handler originated from, so this syntax is mandatory, and also generally preferred in React according to eslint
2. It's a function that has absolutely no side effects: it returns the value purely from transformations, e.g. a function that returns a boolean, maps an array, etc.
3. I need to type the entire function itself, e.g. the case with lambda handler. You "could" do something like this if you really wanted to preserve `function`:
```ts
const lambdaHandler: LambdaHandler = function() {}
```

Since this is backend, taking the latter points into consideration: I feel that they're relatively uncommon and easy to identify, so the precedent I'd set if I had the freedom to do whatever is:

`Use the "function" syntax unless case #2 or case #3 apply, or a similar scenario at your discretion`

At the end of the day, it shouldn't really matter -> maybe just have a "coding guidelines" doc that just lists general pointers that will primarily be enforced by discretion during code reviews. Let me know if I should modify any signatures to conform to your preferences.

api/routes/websoc/schema.ts
bevm0: That's strange, I guess I never ran into that scenario. I went ahead and restored the original schema

ecxyzzy (approved):
> Since this is backend, taking the latter points into consideration: I feel that they're relatively uncommon and easy to identify, so the precedent I'd set if I had the freedom to do whatever is:
> 
> `Use the "function" syntax unless case #2 or case #3 apply, or a similar scenario at your discretion`
> 
> At the end of the day, it shouldn't really matter -> maybe just have a "coding guidelines" doc that just lists general pointers that will primarily be enforced by discretion during code reviews. Let me know if I should modify any signatures to conform to your preferences.

I like this. The main reason I was hesitant is because that would require us to actually have a codified style guide, as we've been getting away with not having one since we have Prettier and ESLint to take care of all of our dirty code. But it's probably good practice to have one somewhere anyway, even if it is "just let the linters do their work except for this one (1) point".

That having been said, this looks good to me. Great work!

PR #25, opened 2023-03-14
feat: implement grades endpoints

ecxyzzy:


api/routes/grades/lib.ts
```diff
@@ -0,0 +1,107 @@
...
+    instructors: { some: { name: { contains: instructor } } },
+    department,
+    courseNumber,
+    courseNumeric,
+    sectionCode,
+    ...excludePNPFilters,
+  };
+}
+
+export function aggregateGrades(grades: GradesRaw): GradesAggregate {
```
Voark, line 66: nit but I feel like this function should have a comment explaning what it does as I think the function name doesn't completely explain it, but maybe I just don't deal with grades data enough to know

Voark (approved):
LGTM

PR #26, opened 2023-03-17
feat: dockerize scraper

ecxyzzy:
TODO:

- [x] Investigate whether it is possible to SSH into the EC2 instance managed by ECS


ecxyzzy:
For the record, it is not possible without significant manual screwing-around (outside of CDK). Since the point of using CDK is to ensure that our infrastructure is replicable, there's not really much point in doing so.

We can however get the container to output logs to CloudWatch, so that's a win in my books.

tools/cdk/app.ts
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env node
 import "dotenv/config";
 
-import { App, StackProps } from "aws-cdk-lib";
+import { type StackProps, App } from "aws-cdk-lib";
```
bevm0, line 4:
```suggestion
import { App } from "aws-cdk-lib";
import type { StackProps } from "aws-cdk-lib";
```
Two reasons that I prefer this (that are mostly trivial)
1. Vim doesn't syntax highlight the mixed line properly for me lol
2. There's actually a prescribed import order by ESLint, and it's easier for me to fulfill them when I just split them up. https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/order.md

ecxyzzy:
Scraper seems to not be reading the previous exit time properly, since it's scraping every term on start. Will do some more testing.

turbo.json
bevm0, line 1: Is there a reason we don't have turborepo depend on all environment variables, and only specific ones?

tools/websoc-scraper-v2/index.ts
```diff
@@ -414,44 +420,41 @@ async function scrape(name: string, term: Term) {
...
-      for (const [name, term] of Object.entries(termsOnDemand)) {
-        if (!(name in termsInScope)) {
-          logger.info(`Term ${name} requested but not in scope, skipping`);
-          continue;
-        }
-        scraped = true;
+    logger.info("websoc-scraper-v2 starting");
+    const prev = new Date(
+      existsSync("/tmp/websoc_scraper_v2")
+        ? parseInt(readFileSync("/tmp/websoc_scraper_v2").toString())
```
bevm0:
```suggestion
        ? parseInt(readFileSync("/tmp/websoc_scraper_v2").toString(), 10)
```
lol.

An alternative that has some overhead, but might bit marginally nicer is `z.coerce.number().parse(...)`, which can parse numeric strings into numbers.

tools/websoc-scraper-v2/index.ts
```diff
@@ -414,44 +420,41 @@ async function scrape(name: string, term: Term) {
...
+        : Date.now() - 24 * 60 * 60 * 1000
+    );
+    const termsInScope = await getTermsToScrape(prev);
+    const curr = new Date();
+    // If it is a new day, scrape all terms.
+    if (curr.getDate() !== prev.getDate()) {
```
bevm0:
```suggestion
    if (curr.getDate() > prev.getDate()) {
```

tools/websoc-scraper-v2/index.ts
```diff
@@ -414,44 +420,41 @@ async function scrape(name: string, term: Term) {
...
         await scrape(name, term);
+    }
+    // Check the database for terms to scrape on demand.
+    const websocTerms = await prisma.websocTerm.findMany({
+      where: { timestamp: { gte: curr } },
+      select: { year: true, quarter: true },
+    });
+    const termsOnDemandEntries = websocTerms.map(
+      (x) => [`${x.year} ${x.quarter}`, x] as TermEntry
+    );
```
bevm0, line 448:
```suggestion
    const termsOnDemandEntries: TermEntry[] = websocTerms
      .map((x) => [`${x.year} ${x.quarter}`, x])
      .filter(([name]) => {
        if (!(name in termsInScope)) {
          logger.info(`Term ${name} requested but not in scope, skipping`);
          return false;
        }
        return true;
    })
```
I think you could actually define the type upfront and then the returned arrays are forced to comply. Also, I think you can remove the `continue` block in the code below just by filtering out the invalid terms early.

tools/websoc-scraper-v2/index.ts
```diff
@@ -414,44 +420,41 @@ async function scrape(name: string, term: Term) {
...
+    const termsOnDemand = Object.fromEntries(termsOnDemandEntries);
+    let scraped = false;
+    for (const [name, term] of Object.entries(termsOnDemand)) {
+      if (!(name in termsInScope)) {
+        logger.info(`Term ${name} requested but not in scope, skipping`);
+        continue;
       }
```
bevm0, line 455:
```suggestion
```
What if we just filter out these cases early?

bevm0 (changes requested):
I think this code is functional; it just needs finetuning to make sure everything is ready for production. Good work! 👍 🔥

tools/websoc-scraper-v2/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM node:16-bullseye-slim
```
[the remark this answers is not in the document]
ecxyzzy: I tried using the Alpine image at first, but it didn't play nice with my laptop's DNS settings (probably due to a quirk in musl libc which it uses). A glibc distro-based container Just Works, and is not that much larger than an Alpine image.

tools/websoc-scraper-v2/build.mjs
```diff
@@ -30,20 +31,15 @@ const dependencies = [".prisma", "db", "registrar-api", "websoc-api-next"];
...
+      .filter((x) => x.includes(".node") && !x.includes("debian-openssl-1.1.x"))
       .map((x) => rm(join(cwd, `${targetDir}node_modules/.prisma/client`, x)))
   );
   await writeFile(
     join(cwd, `${targetDir}.env`),
-    // eslint-disable-next-line turbo/no-undeclared-env-vars
-    `DATABASE_URL="${process.env.DATABASE_URL_SCRAPER}"\nNODE_ENV="production"\nTZ="America/Los_Angeles"`
+    `DATABASE_URL="${process.env.DATABASE_URL_SCRAPER}"\n
+    NODE_ENV="production"\n
+    TZ="America/Los_Angeles"`
```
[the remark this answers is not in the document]
ecxyzzy: If I understand the documentation correctly, Docker recommends against injecting secrets in the `Dockerfile`. And in any case the script already `import`s `dotenv`, so there's not much of a difference IMO.

tools/websoc-scraper-v2/Dockerfile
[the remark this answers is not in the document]
bevm0: Cool! Maybe I'll try using it instead of the vanilla Node image :skull:

turbo.json
bevm0: Oh I see, these env variables don't come from a `.env` file. Ok sure, this sounds good then!

tools/websoc-scraper-v2/build.mjs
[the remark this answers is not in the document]
bevm0: Resolved by setting the container's runtime environment variables in the cdk app.

bevm0 (approved):
The implementation looks solid!

ecxyzzy:
Will hold off on merging this until we have a successful deploy, because the scraper's CloudFormation stack is currently stuck in limbo.

tools/cdk/websocScraperV2Stack.ts
```diff
@@ -1,56 +1,102 @@
...
+     * and enables it for use. Consider removing this in the future
+     * if it becomes unnecessary, e.g. if we get more RAM per instance.
+     */
+    asg.userData.addCommands(
+      "dd if=/dev/zero of=/swap bs=1M count=2048",
+      "chmod 0600 /swap",
+      "mkswap /swap",
+      "swapon /swap",
+      'echo "/swap none swap defaults 0 0" >>/etc/fstab'
+    );
```
bevm0, line 67: I think this is perfectly fine. These exact same commands are being run in the same place where the rest of the container is getting set up.

The issue I had with the shell script was that it seemed out of place in our Node + Docker setup. I felt that those commands should've been encapsulated by either the CDK setup or the Dockerfile

tools/cdk/websocScraperV2Stack.ts
```diff
@@ -1,56 +1,102 @@
...
+      maxCapacity: 1,
       vpc,
-      vpcSubnets: { subnetType: SubnetType.PUBLIC },
     });
-    const dist = new Asset(this, `${id}-dist`, {
-      path: join(
-        dirname(fileURLToPath(import.meta.url)),
-        `../websoc-scraper-v2/dist/websoc-scraper-v2.tar.gz`
-      ),
+    // https://github.com/aws/aws-cdk/issues/18179#issuecomment-1150981559
```
bevm0, line 40:
```suggestion
  /** @see {@link https://github.com/aws/aws-cdk/issues/18179#issuecomment-1150981559} */
```
JSDoc is fun

tools/cdk/websocScraperV2Stack.ts
```diff
@@ -1,56 +1,102 @@
...
+    linuxParameters.addTmpfs({ containerPath: "/tmp", size: 128 });
+    const taskDefinition = new Ec2TaskDefinition(this, `${id}-taskdef`);
+    taskDefinition.addContainer(`${id}-container`, {
+      containerName: `${id}-container`,
+      environment: {
+        DATABASE_URL: process.env.DATABASE_URL_SCRAPER,
+        NODE_ENV: "production",
+        NODE_OPTIONS: "--max-old-space-size=2048",
+        TZ: "America/Los_Angeles",
+      },
```
bevm0, line 86: I noticed there was also an `environment_file` or some similar option. Is there a strict preference or requirement for either? It might be easier to just have the container share the same env file as the project, if that has the same effect.

tools/websoc-scraper-v2/ecosystem.config.js
```diff
@@ -2,7 +2,10 @@ module.exports = {
   apps: [
     {
       name: "websoc-scraper-v2",
-      script: "dotenv -- tsx index.ts",
+      script: "tsx --expose-gc index.ts",
```
bevm0, line 5: I'm not sure if we went over this before, but is there a preference for tsx over ts-node? https://www.npmjs.com/package/ts-node

I usually just use tsx because it seems to work, but ts-node has those juicy npm stars...:eyes:

tools/cdk/websocScraperV2Stack.ts
```diff
@@ -1,56 +1,102 @@
...
+import type { Construct } from "constructs";
 import { dirname, join } from "path";
 import { fileURLToPath } from "url";
 
 export class WebsocScraperV2Stack extends Stack {
   constructor(scope: Construct, id: string, props: StackProps) {
+    if (process.env.NODE_ENV !== "production")
+      throw new Error("Cannot deploy this stack outside of production. Stop.");
+    if (!process.env.DATABASE_URL_SCRAPER)
+      throw new Error("Scraper database URL not provided. Stop.");
```
bevm0, line 29: Is this a special requirement for this cdk script over others? 

While it's definitely good practice to watch out for these conditions, (for better or worse) I usually prefer to avoid adding pre-emptive error checking because it confuses me whether those errors have occurred or not and whether the checking was pre-emptive or reactive.

bevm0 (approved):
I think everything looks fine, great work!👍 

I just left some questions up for discussion.

tools/cdk/websocScraperV2Stack.ts
ecxyzzy: When running in GitHub Actions we won't have an env file, unless we dump the environment into a file in the workflow itself (which feels redundant). In any case the scraper env needs to be distinct from the project env, since it has its own database username/password (for least privilege purposes) and V8 heap options.

tools/websoc-scraper-v2/ecosystem.config.js
ecxyzzy: `tsx` just works, especially since we use ES6 modules exclusively. As you know the corresponding `ts-node` invocation is very verbose.

tools/cdk/websocScraperV2Stack.ts
ecxyzzy: In this case, deploying without the database URL specified will just burn our EC2 time since the deployed container will repeatedly try and fail to connect to the database. So I think it's necessary here IMO.

tools/websoc-scraper-v2/ecosystem.config.js
bevm0: Oh right, I remember the first time you showed me and I was wondering why the CLI for tsx seemed so much nicer lol

PR #30, opened 2023-03-29
feat: update WebSoc endpoint

### LINKED ISSUE #24: Add `cacheOnly` parameter to WebSoc endpoint
When set to `true`, the parameter should make all parameters (except year and quarter) truly optional, and disable falling back to querying WebSoc on an empty response (i.e. just return `{ schools: [] }` if no entries were matched). The default value for this parameter should be `false`. It should be an error if `cache` is set to `false` and this parameter is set to `true`.

### LINKED ISSUE #27: Use cache to send discussions and labs when requesting GEs
When searching for GEs, WebSOC only returns lectures, no co-courses (discussions, labs, etc.), and that's what PeterPortal returns from its endpoint right now. AntAlamanac requests information for each course individually after the initial GE request, resulting in many extraneous requests. 

Now that there's a WebSOC cache, co-courses can be included in the response for GEs. To maintain parity with WebSOC functionality, @ecxyzzy says that a flag should be used to enable this behavior, and lecture-only should be the default.

ecxyzzy:
- Implement support for the `cacheOnly` flag (closes #24).
- Implement support for the `includeCoCourses` flag (closes #27).
- Implement filtering on buildings/rooms, and update the scraper to support this. Somehow the endpoint made it into prod without including this essential feature.

### DIFF
--- a/api/routes/websoc/index.ts
+++ b/api/routes/websoc/index.ts
@@ -77,7 +77,7 @@ export const rawHandler: RawHandler = async (request) => {
                     ? courses[department].push(courseNumber)
                     : (courses[department] = [courseNumber]);
                 });
-                const promises = Object.entries(courses).map(
+                const transactions = Object.entries(courses).map(
                   ([department, courseNumbers]) =>
                     prisma.websocSection.findMany({
                       where: {
@@ -88,7 +88,7 @@ export const rawHandler: RawHandler = async (request) => {
                       distinct: ["year", "quarter", "sectionCode"],
                     })
                 );
-                const responses = (await Promise.all(promises))
+                const responses = (await prisma.$transaction(transactions))
                   .flat()
                   .map((x) => x.data)
                   .filter(notNull) as WebsocAPIResponse[];
--- a/api/routes/websoc/index.ts
+++ b/api/routes/websoc/index.ts
@@ -32,43 +32,85 @@ export const rawHandler: RawHandler = async (request) => {
       try {
         const parsedQuery = QuerySchema.parse(query);
 
-        /**
-         * Check whether an entry with the specified term exists in the sections table.
-         * If not, then we're probably not scraping that term, so just proxy WebSoc.
-         */
-        const termExists = await prisma.websocSection.findFirst({
-          where: { year: parsedQuery.year, quarter: parsedQuery.quarter },
-        });
-
-        if (parsedQuery.cache && termExists) {
-          const websocSections = await prisma.websocSection.findMany({
-            where: constructPrismaQuery(parsedQuery),
-            select: { data: true },
-            distinct: ["year", "quarter", "sectionCode"],
+        if (parsedQuery.cache) {
+          /**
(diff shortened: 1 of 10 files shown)


api/routes/websoc/index.ts
```diff
@@ -32,43 +32,85 @@ export const rawHandler: RawHandler = async (request) => {
...
+               * sections that match from the database.
+               */
+              if (parsedQuery.includeCoCourses) {
+                const courses: Record<string, string[]> = {};
+                websocSections.forEach(({ department, courseNumber }) => {
+                  courses[department]
+                    ? courses[department].push(courseNumber)
+                    : (courses[department] = [courseNumber]);
+                });
+                const promises = Object.entries(courses).map(
```
bevm0, line 80:
```suggestion
                const transactions = Object.entries(courses).map(
```

api/routes/websoc/lib.ts
```diff
@@ -310,6 +310,28 @@ export function constructPrismaQuery(
...
+      AND.push({
+        meetings: {
+          every: {
+            buildings: { some: { bldg: { contains: parsedQuery.building } } },
+          },
+        },
+      });
+    }
+  }
+
```
bevm0, line 334:
```suggestion
  if (parsedQuery.building) {
   AND.push({
        meetings: {
          every: {
            buildings: {
              some: { bldg: parsedQuery.room ? `${parsedQuery.building} ${parsedQuery.room}` : { contains: parsedQuery.building } },
            },
          },
        },
      });
  }
```

libs/websoc-api-next/index.ts
```diff
@@ -239,9 +239,12 @@ export const callWebSocAPI = async (
   json.schools.forEach((school) =>
     school.departments.forEach((department) =>
       department.courses.forEach((course) =>
-        course.sections.forEach(
-          (section) => (section.meetings = getUniqueMeetings(section.meetings))
-        )
+        course.sections.forEach((section) => {
+          section.meetings.forEach(
+            (meeting) => (meeting.bldg = [meeting.bldg] as unknown as string[])
```
bevm0, line 244:
```suggestion
            (meeting) => {
              meeting.bldg = [meeting.bldg].flat()
            }
```
I think this accomplishes the same thing, but without asserting

bevm0 (changes requested):
Good stuff :rocket:

tools/websoc-scraper-v2/tsconfig.json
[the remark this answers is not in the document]
ecxyzzy: Most of our tooling (prettier, typescript-eslint, etc.) should already be compatible with TS5.0, it's just a matter of updating them, which would be out of scope for this PR.

Is there anything in particular that you think would be better addressed by extending multiple tsconfigs?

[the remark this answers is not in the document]
bevm0: no, not really unless you felt that you needed it; using those features breaks some of the tools I personally use

bevm0 (approved):
Epic work :fire:

Issue #17, opened 2023-03-04
WebSoc terms list REST endpoint


ecxyzzy:
Implement the WebSoc terms list REST endpoint (`/v1/rest/websoc/terms`). It shall accept no additional parameters.

The payload of a successful response from this endpoint shall match the type that is outlined in the types package [here](https://github.com/icssc/peterportal-api-next/blob/main/packages/peterportal-api-next-types/types/websoc.ts#L226-L254).

Issue #16, opened 2023-03-04
WebSoc departments list REST endpoint


ecxyzzy:
Implement the WebSoc departments list REST endpoint (`/v1/rest/websoc/departments`). It shall accept no additional parameters.

The payload of a successful response from this endpoint shall match the type that is outlined in the types package [here](https://github.com/icssc/peterportal-api-next/blob/main/packages/peterportal-api-next-types/types/websoc.ts#L211-L224).

PR #34, opened 2023-05-03
feat: add websoc depts and grades endpoints

y-dejong:
Added WebSoc depts and grades endpoints under /v1/websoc/:options.
Solves #16 and #17


ecxyzzy (changes requested):
Response formats look good!

One small nit—the departments and terms from the grades db seem like they're out of order. Could you make it so that the responses are sorted in department code alphabetical and reverse chronological order respectively?

ecxyzzy (approved):
lgtm now, great work!

PR #44, opened 2023-05-26
Update types package

Voark:
Updated types package to 1.0.0-beta.1 so it can be published to npm.

I also published the package to npm.


ecxyzzy (approved):
ɯʇƃl

PR #45, opened 2023-05-27
feat: test api performance optimizations

ecxyzzy:


ecxyzzy:
Doubling the lambda memory has reduced execution time in some cases, but did not resolve unexplained long invocations (probably due to Prisma). More detailed profiling is necessary but I don't have the time for it now and we'll be merging in DX changes soon™, so I'll be closing this as outdated for now.

PR #46, opened 2023-05-30
feat: pnpm and monorepo patches

ap0nia:
Added pnpm and did some project fixes.

- Changed config files to `.cjs` to accommodate type checking with JSDoc
- `Prisma` seems to interact with the `pnpm` workspace slightly differently from `npm`, so I updated the build scripts to accommodate that.
- Also made the build scripts more uniform. My next PR will abstract all the build logic in a dedicated package.
- There's a conflicting package on npm called [`api-core`](https://www.npmjs.com/package/api-core), so I needed to specify `workspace:*` in the version for the local package. Just for consistency, I also did that for all local packages.
- `pnpm` properly scopes dependencies for each sub-project, so I needed to adjust some `package.json`s to make sure they contained all the dependencies they needed.


.eslintrc.cjs
```diff
@@ -0,0 +1,44 @@
+/** @type {import('eslint').Linter.Config} */
```
ap0nia, line 1: Using JSDoc + JS config formats can actually give you type information for the ESLint and Prettier configurations, which is pretty neat. Otherwise, the settings should be the same.

.prettierrc.cjs
```diff
@@ -0,0 +1,9 @@
+/**
+ * `cjs` format is required for plugins to work in VSCode.
+ * @see {@link https://github.com/tailwindlabs/prettier-plugin-tailwindcss/issues/113}
+ */
+
```
ap0nia, line 5: I left this note around when dealing with a full stack application w/TailwindCSS. It's up to your discretion whether it's worth leaving around.

api/cdk/cdk.json
```diff
@@ -1,5 +1,5 @@
 {
-  "app": "tsx app.ts",
+  "app": "npx tsx app.ts",
```
ap0nia, line 2: When invoking `cdk` from the terminal CLI directly, it doesn't have the context of node modules. Adding `npx` should allow the command to work when invoked from the terminal or with an npm command.

ecxyzzy: Why not `pnpx` in this case? Genuine question, since I'm not as familiar as I'd like with the `pnpm` ecosystem.

package.json
```diff
@@ -2,25 +2,50 @@
...
     "deploy": "turbo run deploy",
     "destroy": "turbo run destroy",
-    "dev": "dotenv -c development -- turbo run dev",
-    "format": "prettier --write .",
-    "generate": "turbo run generate",
-    "postinstall": "husky install",
+    "commit": "cz",
     "lint": "eslint --fix .",
-    "test": "turbo run test"
+    "format": "prettier --write ."
```
ecxyzzy, line 14: Was there a reason for eliminating the global test/dev scripts? I know we don't have many tests at the moment but that is something I would like to change in the future, and also being able to run `npm run dev` from the root is great for testing e.g. the GraphQL endpoint using Apollo Sandbox, which is on the docs site.

ecxyzzy (changes requested):
See comments.

Also I probably won't re-review until after finals since the consequences of my actions caught up to me, so no need to rush on responding.

package.json
ap0nia: Oh oops, I thought each sub-project had its own dev script; that's how I setup the routes with the custom framework.

I just removed test since I didn't see anything there, but I guess I'll just add it back :^)

api/cdk/cdk.json
ap0nia: The main reason I've pondered over is that you're guaranteed to have npx because that's built in, but not always pnpx. But you should have pnpx if you're working on the project :shrug: 

I guess there's also the case that using npx everywhere is more consistent; the CLIs of the package managers can behave slightly differently for some reason, so maybe using a well-established one might be more predictable

ecxyzzy:
tfw you finally fix the build system and the message doesn't even render properly :upside_down_face:

ecxyzzy (approved):
LGTM now, excellent job and thanks for working with me on this!

PR #47, opened 2023-06-01
feat: PeterPortal API SST

ap0nia:
Built on top of #46 . Will point this PR to that branch temporarily to reduce the number of files changed.

Does ***not*** actively affect the project. The package and its related tools are simply added to the project. They can be used to develop new endpoints.

Mainly missing built-in CDK support. I want to add a CLI script that will automatically read all projects in a specified directory, and deploy them all as separate AWS Lambda functions. [The relevant stack and run scripts have already been created here](https://github.com/icssc/peterportal-api-next/tree/peterportal-api-sst/packages/peterportal-api-sst/src/cdk), but I'm not sure how to bundle it into the package.

- Also fixes some issues with docusaurus types.


env.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
+
+/**
+ * ESbuild will pick up on this and copy the env file to the output folder.
+ */
+// eslint-disable-next-line
+try {
+  require("./.env");
+} catch {}
```
ecxyzzy, line 15:
```suggestion
/**
 * esbuild will pick up on this and copy the env file to the output folder.
 */
try {
  require("./.env");
} catch { /* noop */ }
```
This actually works and is IMO more concise than using eslint-disable macros.

packages/peterportal-api-sst/src/cdk/index.ts
```diff
@@ -0,0 +1,69 @@
+import { existsSync, readdirSync } from "node:fs";
+import { relative } from "node:path";
+
+import * as cdk from "aws-cdk-lib";
+
+import { getConfig } from "../config.js";
+import { type HandlerConfig, PeterPortalAPI_SST_Stack } from "./stack.js";
+
+// import { env } from '../../../env'
+let env: any;
```
ecxyzzy, line 10:
```suggestion
let env: Record<string, string> | undefined;
```

packages/peterportal-api-sst/src/cdk/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+  const subRoutes = readdirSync(`${apiDir}/${route}`);
+  return subRoutes.flatMap((subRoute) => getApiRoutes(`${route}/${subRoute}`, apiDir, current));
+};
+
+async function start() {
+  const config = await getConfig();
+
+  config.env = env ?? {};
+  config.env.stage = getStage(env.NODE_ENV);
```
ecxyzzy, line 45:
```suggestion
  config.env.stage = getStage(env?.NODE_ENV);
```

packages/peterportal-api-sst/src/cdk/index.ts
```diff
@@ -0,0 +1,69 @@
+import { existsSync, readdirSync } from "node:fs";
+import { relative } from "node:path";
+
+import * as cdk from "aws-cdk-lib";
+
+import { getConfig } from "../config.js";
+import { type HandlerConfig, PeterPortalAPI_SST_Stack } from "./stack.js";
```
ecxyzzy, line 7: Class names should always use PascalCase, never snake_case.

Assuming you stick with the SST nomenclature (see the main review comment) an acceptable name might be `ApiSstStack`. See also [here](https://google.github.io/styleguide/tsguide.html#identifiers-abbreviations).

packages/peterportal-api-sst/src/cdk/stack.ts
```diff
@@ -0,0 +1,155 @@
+import * as cdk from "aws-cdk-lib";
+import type { Construct } from "constructs";
+
+import type { PPA_SST_Config } from "../config.js";
```
ecxyzzy, line 4: Type names should also be PascalCase.

packages/peterportal-api-sst/src/lambda-core/constants.ts
```diff
@@ -0,0 +1,39 @@
...
+  "Nov",
+  "Dec",
+] as const;
+
+export const httpMethods = ["DELETE", "GET", "HEAD", "PATCH", "POST", "PUT", "OPTIONS"] as const;
+
+export type HttpMethod = (typeof httpMethods)[number];
+
+export function isHttpMethod(method: string): method is HttpMethod {
+  return httpMethods.includes(method as any);
```
ecxyzzy, line 38:
```suggestion
  return httpMethods.includes(method as (typeof httpMethods)[number]);
```

ecxyzzy (changes requested):
See comments.

Some additional issues:
- Very small nit, but maybe consider an alternative name since this framework is not the same as the [SST](https://sst.dev/) you told me about.
- Is there a reason to keep the Bun code around? I don't really see us migrating to it any time soon, even though it looks very interesting.

packages/peterportal-api-sst/src/cdk/index.ts
ap0nia: Oh, I forgot to remove the `env` stuff; it should always be referring to `config.env`, which I already ensured was defined.

packages/peterportal-api-sst/src/lambda-core/constants.ts
ap0nia: I think I prefer typing it as `any` because that's essentially what the input string should be -- any string. But TS gets upset that I'm passing in an arbitrary string so I just wanna turn off the type checking.

ap0nia:
Idk what to name it, give me suggestions!

I understand the JS camelCase convention, but I think I'd like to finalize a name for the package first. Basically all variable names I've tried look kinda goofy because the framework's name is composed entirely of acronyms lol. Once we have a name we're happy with, I'll go fix all the type names :^)

The framework will have more features if it supports Bun! It was mostly implemented as a proof of concept, and so I can test out Bun integration with AWS Lambda. Honestly, I think results are kinda inconclusive, but I'm fine with keeping it around so I can play around with it more in the future.

packages/peterportal-api-sst/src/lambda-core/constants.ts
ecxyzzy: Yeah I guess so, but eslint doesn't like it. Could you add a eslint-disable macro for this then?

ecxyzzy:
> Idk what to name it, give me suggestions!
> 
> I understand the JS camelCase convention, but I think I'd like to finalize a name for the package first. Basically all variable names I've tried look kinda goofy because the framework's name is composed entirely of acronyms lol. Once we have a name we're happy with, I'll go fix all the type names :^)

`AntStack`? It goes well with the UCI theme, and also it does all of the heavy lifting for the API, because ants have remarkable organizational skills and impressive strength-to-weight ratios :muscle: 

> The framework will have more features if it supports Bun! It was mostly implemented as a proof of concept, and so I can test out Bun integration with AWS Lambda. Honestly, I think results are kinda inconclusive, but I'm fine with keeping it around so I can play around with it more in the future.

Okay makes sense. Just wondering, but is there a way to opt out of building the Bun stuff for production endpoints if we end up not using it?

ap0nia:
Sounds like we want to enhance the config! Which we can do if we want.

The code that "builds" the lambda just writes a couple of lines to a file lol.
https://github.com/icssc/peterportal-api-next/blob/peterportal-api-sst/packages/ant-stack/src/cli/commands/build.ts#L56

The idea is that the framework compiles a runtime agnostic handler function, and we'd have to generate additional files to plug in the adapters for the specific runtime we want to target. So the end of the build script just does that. There's technically no "building the lambda runtime"; but we can tune the config to allow different cases!

ecxyzzy (approved):
LGTM now, excellent work! Excited to see how much better the DX will be going forward :rocket:

I guess it's a good thing that I haven't built the instructors endpoint yet and have the whole afternoon to test it out :eyes:

PR #48, opened 2023-06-17
feat(ant-stack): implement route creation logic

ecxyzzy:
Notable changes:
- Removed the background colors because the text with them were either illegible or made my eyes bleed. Generally I prefer minimal use of colors, even for a CLI tool, like how `pnpm` does it (as shown below).
image
- Removed non-interactive route creation. We can add it back in later, but I feel that a tool like this should focus on DX, which means prioritizing interactivity.
- Removed `DATABASE_URL_SCRAPER` from the env. Normal members are not going to have access to it, and they don't need it for API route development anyways, which is what AntStack is for.

Caveats:
- I'm not sure how to start the development server properly; it errors at the root, and serves the given endpoint as the root when I run it from the endpoint's directory. Neither of these behaviors are desirable.

### DIFF
--- a/env.ts
+++ b/env.ts
@@ -4,8 +4,6 @@ import { fileURLToPath } from "node:url";
 import { type } from "arktype";
 import { config } from "dotenv";
 
-// eslint-disable-next-line @typescript-eslint/ban-ts-comment
-// @ts-ignore
 const __dirname = fileURLToPath(new URL(".", import.meta.url));
 
 /**
--- a/packages/ant-stack/src/cli/commands/create.ts
+++ b/packages/ant-stack/src/cli/commands/create.ts
@@ -28,7 +28,9 @@ export async function interactiveCreate() {
     path = await consola.prompt("What is the path of the endpoint?", { type: "text" });
     if (!(path.match(/\/[0-9A-Za-z]+/) && !path.endsWith("/"))) {
       consola.error(
-        "Malformed path provided. A well-formed path must consist entirely of path parts (one slash followed by at least one alphanumeric character), and must not end with a slash (e.g. /v1/rest/test)."
+        chalk.red(
+          "Malformed path provided. A well-formed path must consist entirely of path parts (one slash followed by at least one alphanumeric character), and must not end with a slash (e.g. /v1/rest/test)."
+        )
       );
       continue;
     }
@@ -42,7 +44,7 @@ export async function interactiveCreate() {
       { type: "confirm" }
     );
     if (!create) {
-      consola.error("Aborting.");
+      consola.error(chalk.red("Aborting."));
       return;
     }
   }
--- a/packages/ant-stack/src/cli/commands/create.ts
+++ b/packages/ant-stack/src/cli/commands/create.ts
@@ -23,7 +23,7 @@ export async function interactiveCreate() {
   const config = await getConfig();
   consola.info(chalk("Creating a new endpoint."));
   let path = "",
-    success = false;
+  let success = false;
   while (!success) {
     path = await consola.prompt("What is the path of the endpoint?", { type: "text" });
     if (!(path.match(/\/[0-9A-Za-z]+/) && !path.endsWith("/"))) {
--- a/packages/ant-stack/src/cli/commands/create/index.template.ts
+++ b/packages/ant-stack/src/cli/commands/create/index.template.ts
@@ -1,0 +1,16 @@
+import { createErrorResult, createOKResult, type InternalHandler, zeroUUID } from "ant-stack";
+
+const GET: InternalHandler = async (event) => {
+  return createOKResult({}, zeroUUID);
+};
+const POST: InternalHandler = async (event) => {
+  return createOKResult({}, zeroUUID);
+};
+const PUT: InternalHandler = async (event) => {
+  return createOKResult({}, zeroUUID);
+};
+const DELETE: InternalHandler = async (event) => {
+  return createOKResult({}, zeroUUID);
+};
+
+export default {};
--- a/packages/ant-stack/src/cli/commands/create/index.ts
+++ b/packages/ant-stack/src/cli/commands/create/index.ts
@@ -1,18 +1,13 @@
 import chalk from "chalk";
 import { consola } from "consola";
-import { existsSync, mkdirSync, writeFileSync } from "fs";
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from "fs";
 import { join } from "path";
 import { format } from "prettier";
+import { fileURLToPath } from "url";
 
 import { getConfig } from "../../../config.js";
 
-const packageJson = (name: string) =>
-  `{"name": "api-${name
-    .slice(1)
-    .replace(
-      /\//g,
-      "-"
-    )}","version": "0.0.0","type": "module","scripts":{"dev": "ant dev","build": "ant build","start": "node dist"},"dependencies": {"ant-stack": "*"},"devDependencies":{"@types/node": "18","tsx": "*","typescript": "~5.0"}}`;
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
 
 const imports = `import { createErrorResult, createOKResult, type InternalHandler, zeroUUID } from "ant-stack"\n\n`;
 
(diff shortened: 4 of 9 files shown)


ap0nia:
- colors: I just wanted to use all the colors to see what they looked like lol. They were definitely very goofy!
- interactivity: valid; a working CLI is a better priority.
- env: sure, that stuff is configurable anyways.

- Oh, that's actually how I programmed the dev server; it'll spin up an express server for the particular endpoint with the handler responding to requests to the root. It's just a really quick way to test your own endpoint individually. I haven't configured it to be able to run at the root. To run at the root, ESBuild needs to run on every endpoint individually, watch all the subprojects, and properly assign routes to all the handlers whenever it reloads. I think it's a lot of stuff, do you think it's high priority for this tool?

packages/ant-stack/src/cli/commands/create.ts
```diff
@@ -1,24 +1,69 @@
...
+    .slice(1)
+    .replace(
+      /\//g,
+      "-"
+    )}","version": "0.0.0","type": "module","scripts":{"dev": "ant dev","build": "ant build","start": "node dist"},"dependencies": {"ant-stack": "*"},"devDependencies":{"@types/node": "18","tsx": "*","typescript": "~5.0"}}`;
+
+const imports = `import { createErrorResult, createOKResult, type InternalHandler, zeroUUID } from "ant-stack"\n\n`;
+
+const handlers = (method: string) =>
+  `export const ${method}: InternalHandler = async (event) => {return createOKResult({}, zeroUUID);};`;
```
ap0nia, line 20: Instead of generating the strings from scratch, can we use a template file and then replace the parts as needed. Here's an example of a template file from Svelte: https://github.com/sveltejs/kit/blob/master/packages/create-svelte/templates/default/package.template.json

The stuff they'll replace later is annotated as `~TODO~`; we can create any number of arbitrary symbols to replace per the user's desired options.

packages/ant-stack/src/cli/commands/create.ts
```diff
@@ -1,24 +1,69 @@
...
 
 export async function interactiveCreate() {
-  consola.info(chalk.blue("Create a new endpoint interactively!"));
-  await consola.prompt("What is the name of the endpoint?", { type: "text" });
-  await consola.prompt("What HTTP methods does it use?", {
+  const config = await getConfig();
+  consola.info(chalk("Creating a new endpoint."));
+  let path = "",
+    success = false;
```
ap0nia, line 26:
```suggestion
  let path = "",
  let success = false;
```
I don't really like multiple assignments on the same line because it's weird to my smol brain :pensive:

packages/ant-stack/src/cli/commands/create.ts
```diff
@@ -1,24 +1,69 @@
...
-  consola.info(chalk.greenBright.bgRedBright(options));
-  consola.info(chalk.green.bgYellow(`Endpoint created!`));
+  consola.info(`Creating an endpoint at ${path} that supports ${methods.join(", ")}`);
+  mkdirSync(srcDir, { recursive: true });
+  writeFileSync(join(srcDir, "..", "package.json"), format(packageJson(path), { parser: "json" }));
+  writeFileSync(
+    join(srcDir, "index.ts"),
+    format([imports, ...methods.map((method) => handlers(method))].join(""), {
+      parser: "babel",
+    })
```
ap0nia, line 62:
```suggestion
```
If we use template files, they should be formatted properly and we'd just replace whatever text we need to.

packages/ant-stack/src/config.ts
```diff
@@ -59,6 +59,11 @@ interface AntRuntime {
  * AntStack's development server configuration.
  */
 export interface AntConfig {
+  /**
+   * The package manager used by the project.
+   */
+  packageManager: "npm" | "yarn" | "pnpm";
```
ap0nia, line 65: In this config file, I use [`defu`](https://github.com/unjs/defu) to recursively assign default values. I think we can do something like "look for workspace root, and figure out what kind of lock file is being used", to figure out what this is. i.e. it'll be based on whether the lock file is a package-lock.json, yarn.lock, or pnpm-lock.yaml

[Vite came up with a way to find the workspace root if we'd want to consider something like this](https://github.com/vitejs/vite/blob/711dd807610b39538e9955970145d52e4ca1d8c0/packages/vite/src/node/server/searchRoot.ts#L46)

ecxyzzy:
This would definitely be high priority, since the GraphQL endpoint depends on all of the other endpoints. So while it's fine for every other endpoint to be tested individually this way, we wouldn't be able to test the GraphQL endpoint at all, apart from requesting a staging deployment and testing it from there.

env.ts
```diff
@@ -4,6 +4,8 @@ import { fileURLToPath } from "node:url";
 import { type } from "arktype";
 import { config } from "dotenv";
 
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
 const __dirname = fileURLToPath(new URL(".", import.meta.url));
```
ap0nia, line 9: Is there something wrong with this? If it's just the name of the variable, we could change it to `_dirname_` or something just to satisfy TS lol.

ap0nia (changes requested):
Thanks for helping! I think the functionality in the create command is pretty solid; I'd like to try using template files to see if that's any better

env.ts
ecxyzzy: To be honest I'm not entirely sure why this happens, might be my IDE being stupid since I copypasted this exact line into another file under the same tsconfig scope and it didn't error at all. I'll remove it.

packages/ant-stack/src/config.ts
ecxyzzy: Honestly this feels like more trouble than it's worth, and most other `create-<x>-app` tools ask you what package manager you want for the project anyways. Since we'll almost certainly want to write a create/bootstrap script, I don't think it's cheating to just get it from the initial user input and store it in our config file.

Besides, surely no one would want to migrate package managers after already creating their project, right…? :zipper_mouth_face:

ap0nia: It's ok, I implemented the core logic on my dev stuff ... skill diff :sunglasses:

packages/ant-stack/src/cli/index.ts
```diff
@@ -1,14 +1,13 @@
...
-import { defu } from "defu";
 
 import { buildInternalHandler } from "./commands/build.js";
-import { automaticCreate, CreateOptions, interactiveCreate } from "./commands/create.js";
+import { interactiveCreate } from "./commands/create";
 import { startDevServer } from "./commands/dev.js";
 
 async function start() {
-  consola.log(chalk.green.bgRedBright(`ant - AntStack's CLI tool`));
+  consola.log(chalk("🐜 ant :: the AntStack command-line tool"));
```
ap0nia: Nice touch :sunglasses:

packages/ant-stack/src/cli/index.ts
```diff
@@ -70,5 +47,5 @@ async function start() {
 start();
 
 export * from "./commands/build.js";
-export * from "./commands/create.js";
+export * from "./commands/create";
 export * from "./commands/dev.js";
```
ap0nia, line 51: I think we can remove this actually, because this file literally just executes. The only reason for imports/exports is to get the types, but you can just navigate directly to the file I'd imagine.

packages/ant-stack/src/cli/commands/create/index.template.ts
ap0nia, line 1: Ok sorry, I was the one that suggested the templates, but I think dynamically generating the entry file makes sense. We can leave this here as a reference tho.

I think we can do something like this:
```ts
function createHandlerTemplate(httpMethod: string) {
  return `\
const ${httpMethod}: InternalHandler = async (event) => {
  return createOKResult({}, zeroUUID);
};
`
}
const fileTemplate = methods.map(createHandlerTemplate).join('')
fs.writeFileSync(...)
```
What do you think? It's mostly because keeping track of file lines can get tricky if we ever update the template file, and the entry file is very repetitive, unlike the package.json.

packages/ant-stack/src/cli/commands/create/index.ts
```diff
@@ -0,0 +1,71 @@
...
+  );
+  const indexLines: (string | null)[] = readFileSync(
+    join(__dirname, "../src/cli/commands/create/index.template.ts"),
+    { encoding: "utf-8" }
+  ).split("\n");
+  for (const method of Object.keys(methodLines)) {
+    if (!methods.includes(method)) {
+      const i = methodLines[method];
+      indexLines[i] = indexLines[i + 1] = indexLines[i + 2] = null;
+    }
```
ap0nia, line 62: I think we can try the strat I mentioned above with dynamically generating this particular file.

ap0nia (changes requested):
Ok, just one final thing: revising my previous suggestion, I think it does make sense for the entry file to be dynamically generated because it's very repetitive. But we can keep the template.ts file around as a reference to how it should look like.

ecxyzzy (commented):
GitHub machine :b:roke and left me with a ghost review, so I'll just submit this to get rid of it. Nothing to see here, move along.

ap0nia (approved):
Looks good! Great work :pray:

PR #51, opened 2023-06-18
feat: goofy-ahh emojis

ap0nia:
Very important feature, framework doesn't work without it


ecxyzzy (approved):
Agreed, nice catch. ship it :boat:

PR #53, opened 2023-06-18
chore: Workspace cleanup

ap0nia:
really massive codebase cleanup lol

### Caveat
- I removed the "hoist" from the .npmrc in the project root because I have time to fix whatever issues arise with building. Let me know if anything pops up

### Files Changed ...
Should ignore the project files, like `.ts` because those were just formatted.

- .commitlintrc.json - removed because the same config is already in package json
- .lintstagedrc.yml - extracted out from the root package.json
- docs --> apps/documentation
- .eslintrc.cjs --> simplified config, also added support for react because of the docs
- .pretterrc.cjs --> prettier.config.cjs (for some reason this specific config file is better with plugins)
- tsconfig.json --> edited

- docusaurus.config.js --> docusaurus.config.cjs
- sidebars.js --> sidebars.cjs
- babel.config.js --> babel.config.cjs
- changing these files in the documentation project don't seem to affect functionality, but it helps fix linting/TS errors complaining that they should be modules

Other miscellaneous notes:
- I generally prefer config stuff in dedicated config files instead of package json for consistence with hooks and stuff. But commitizen and commitlint seem really similar to me, and I'm fine with their configs in package.json . Personal, inconsistent preferences
- removed "license MIT" from sub-projects, idk I feel like nobody looks at them so I strip out everything that's not essential lol
- did you know `.cts` and `.mts` are the equivalent TS file extensions for `.cjs` and `.mjs` ? It looks kinda goofy and I've rarely ever seen them used


ap0nia:
Ok, will make a new PR to this branch specifically. Will turn off commit hooks and only submit the config changes for review. Then will do a project-wide lint and format after everything looks good.

PR #56, opened 2023-06-21
ci: move typechecking to GitHub Actions

ecxyzzy:


.github/workflows/check-typescript.yml
```diff
@@ -0,0 +1,9 @@
+name: Check for TypeScript errors
```
ap0nia, line 1:
```suggestion
name: Test
```
Maybe we can include the test suite along with the type checking in the future, or maybe we can always keep them separate so they run in parallel? Just some ideas

ap0nia (approved):
Simple change, looks solid!

## More Considerations for linting/formatting and hooks
[Theo says that we **shouldn't** use pre-commit hooks because it slows down the developer workflow.](https://www.youtube.com/watch?v=RAelLqnnOp0)

My first interpretation of this was that the CI/CD server should do the formatting/linting and then commit automatically. I tried this, and this was absolutely horrendous because it would de-sync the local branch; it was a purely inferior version of pre-commit hooks.

The correct interpretation of this is that the CI/CD server will run the formatting and linting ___without modifying any files___ and then just report anything not within spec. This prevents merging, but doesn't stop the developer from continuing their work. When they're ready for a review, the developer should run formatting, linting, etc. to ensure the CI/CD server doesn't report any errors.

My opinion: pre-commit hooks for linting/formatting seems really trivial, and type checking, testing, etc. can be reserved for dedicated remote servers. This system seems to help me stay focused on immediately important tasks instead of remembering when to do x, y, z.

.github/workflows/check-typescript.yml
ecxyzzy: Let's keep them separate for now. We can also separate running tests into its own action down the line instead of having it be part of the staging action.

PR #58, opened 2023-06-29
feat: release v1.0.0-rc.0

ecxyzzy:


ap0nia (approved):
My brain is fried and I don't have the capacity for insightful feedback, sorry. 😵‍💫
I think it looks fine :^)

.github/workflows/cleanup-pr-api.yml
```diff
@@ -72,18 +64,20 @@ jobs:
...
-      - name: Delete staging URLs
-        uses: marocchino/sticky-pull-request-comment@v2
+      - name: Delete environment
+        if: github.event.action == 'closed'
+        uses: strumwolf/delete-deployment-environment@v2.3.0
         with:
-          header: stagingURL
-          delete: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          environment: staging-${{ github.event.pull_request.number }}
```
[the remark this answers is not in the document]
ecxyzzy: Nah. If it ain't broke, don't fix it™

.husky/pre-commit
```diff
@@ -1,11 +1,4 @@
 #!/usr/bin/env bash
```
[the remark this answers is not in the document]
ecxyzzy: It's not necessarily the case that `/bin/sh -> bash`. On Debian (and downstream) I believe it links to `dash`, which is not `bash`-compatible. I checked our `pre-commit` and `commit-msg` scripts and all of them have `bash`-isms, so I'm keeping the shebang as is.

Husky's init script (which is ignored) seems to be free of `bash`-isms, so it uses `#!/usr/bin/env sh`.

apps/api/v1/rest/grades/src/index.ts
```diff
@@ -0,0 +1,125 @@
...
+import type { InternalHandler } from "ant-stack";
+import type { GradesOptions, GradesRaw } from "peterportal-api-next-types";
+import { ZodError } from "zod";
+
+import { aggregateGrades, constructPrismaQuery, lexOrd } from "./lib";
+import { QuerySchema } from "./schema";
+
+let prisma: PrismaClient;
+
+export const GET: InternalHandler = async (request) => {
```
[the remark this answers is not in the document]
ecxyzzy: Oh right I forgot to mention. There's no more need for this, because Express handles `HEAD` requests, and I modified the cdk code to also handle it in staging/prod if any route has a `GET` request.

The old solution meant creating a carbon copy of the `GET` handler which would also have its own DB connections. You can see how that would be unsustainable.

package.json
```diff
@@ -5,10 +5,9 @@
   "description": "The official PeterPortal API-Next monorepo",
   "type": "module",
   "scripts": {
-    "ant:build": "dotenv -c development -- ant build",
-    "ant:dev": "dotenv -c development -- ant dev",
     "build": "turbo run build",
     "commit": "cz",
+    "create": "dotenv -v DATABASE_URL=null -v NODE_ENV=development -- pnpm exec ant-stack create",
```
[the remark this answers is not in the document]
ap0nia: [`cross-env`](https://www.npmjs.com/package/cross-env) is meant to address cross-compatibility between different shells. Idk why shells aren't just all POSIX compliant (cough, Windows) or set environment variables the same, but skill diff :shrug:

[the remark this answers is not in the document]
ecxyzzy: So does `dotenv-cli`:
image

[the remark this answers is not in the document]
ap0nia: Oh ok, illiteracy moment. It's just an npm popularity diff then

apps/api/v1/rest/calendar/ant.config.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { cleanCopy, selectDelete } from "@libs/build-tools";
+
+// ESM hack for __dirname
+const cwd = dirname(fileURLToPath(import.meta.url));
+
+// The relative path to the generated Prisma Client.
+const prismaClientDir = "./node_modules/@libs/db/node_modules/prisma";
+
+const prismaSchema = "./node_modules/@libs/db/prisma/schema.prisma";
```
ap0nia, line 15: I don't think it's worth changing (especially with new devops stuff coming out soon :tm:), but I personally like to use something like the `getNearestProjectDirectory` utility to ensure a consistent starting directory, and then using `path.resolve` with each path segment individually so they're chained in a normalized fashion.

e.g.
```ts
const prismaSchema = path.resolve(getProjectDirectory(), 'node_modules', '@libs', 'db', 'node_modules', 'prisma')
```

It's just something I found to be helpful when it actually matters whether or not there's random duplicated slashes, like `/node_modules//@libs/db`, or I just want to know exactly where I'm starting at.

apps/api/v1/rest/websoc/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+  combineResponses,
+  constructPrismaQuery,
+  normalizeQuery,
+  notNull,
+  PeterPortalApiLambdaClient,
+  sortResponse,
+} from "./lib";
+import { QuerySchema } from "./schema";
+
+let prisma: PrismaClient;
```
ap0nia, line 17: I don't think this includes the current implementation of `ant-stack` in use for this PR, but the latest iteration of the framework will do a static analysis of the file without executing it, so it will be perfectly fine to initialize prisma here as usual (once I get the chance to release it).

apps/api/v1/rest/websoc/src/index.ts
```diff
@@ -0,0 +1,222 @@
...
+
+const lambda = new LambdaClient({});
```
ap0nia, line 19: If it's possible to initialize the `LambdaClient` without any meaningful options, then I think the class should provide a default argument.

```suggestion
const lambda = new LambdaClient();
```

lint-staged.config.cjs
ap0nia, line 1: JSDoc config is generally really neat, but the types for `lint-staged` are useless lol. It's basically `Record<string, string[]>`.

YAML is nicer to read at this point tbh, but I do like the "just do all the config files in JS with JSDoc" convention because it keeps things a lot more predictable/consistent :^)

ap0nia (approved):
Cool, thanks for taking a look at some of my previous suggestions :pray:

apps/api/v1/rest/websoc/src/index.ts
ecxyzzy: Yeah unfortunately you have to pass in the empty object. AWS SDK PR when?

lint-staged.config.cjs
ecxyzzy: Be the change you wish to see in the world :^)

I mean yeah, at the end of the day that is effectively all the configuration that is available for lint-staged. I just like it because consistency.

apps/api/v1/rest/websoc/src/index.ts
ap0nia: Oh mb, I thought this was the client that we created lol. I remember one that we used internally for making requests, but I forgot where it was

PR #68, opened 2023-08-01
feat!: 💥 ✨ normalize section meeting and final times (WIP)

### LINKED ISSUE #72: Normalize section meeting and final exam times

### LINKED ISSUE #92: Fallback to WebSoc proxy if Prisma can't get a database connection
Instead of crashing and not responding at all.

ecxyzzy:
~~Currently very slow even from cache, will be investigating why.~~

Disregard, local dev server diff. Requests are still blazingly fast :rocket:

Closes #72; closes #92.


ecxyzzy:
Looks like icssc/AntAlmanac#669 already got merged so I guess this is ready for review.

apps/api/v1/graphql/src/graphql/enum.graphql
```diff
@@ -55,10 +55,16 @@ enum CancelledCourses {
   Include
   Only
 }
-"The set of valid options for filtering based on enrollment status."
+"The set of valid enrollment statuses."
```
ap0nia, line 58: Based on the comment for the "WebsocSectionFinalExamStatus" enum, maybe this one should be "The status of an enrollment" or something similar. Or, the websoc enum can be more like "The set of valid final exam statuses from websoc". I like the first format more.

apps/api/v1/graphql/src/graphql/websoc.graphql
```diff
@@ -1,11 +1,27 @@
...
 type WebsocSectionMeeting {
-  "What day(s) the section meets on (e.g. ``MWF``)."
-  days: String!
-  "What time the section meets at."
-  time: String!
-  "The building(s) the section meets in."
+  """
+  Whether the meeting time is TBA.
+
+  If this field is `false`, then `days`, `startTime`, and `endTime`
```
ap0nia, line 13: I don't think there's any significant LSP behavior that GraphQL offers, but if I wanted to refer to other properties, types, or objects in JSDoc, I'd use the `@link` indicator. 

```ts
interface Foo {
  bar: string

 /**
  * Should be the same as {@link bar}
  */
  baz: string
}


```

apps/api/v1/rest/websoc/ant.config.ts
```diff
@@ -18,6 +18,7 @@ const outDir = resolve(cwd, "./dist");
 
 const config: AntConfigStub = {
   esbuild: {
+    external: process.env.NODE_ENV === "development" ? [] : ["@services/websoc-proxy-service"],
```
ap0nia, line 21: Does the dev server break or something if we try to bundle it during development?

apps/api/v1/rest/websoc/package.json
```diff
@@ -9,11 +9,13 @@
   "dependencies": {
     "@aws-sdk/client-lambda": "3.388.0",
     "@libs/db": "workspace:*",
+    "@libs/websoc-utils": "workspace:*",
```
ap0nia, line 12: Not a big deal, and maybe we can do this during a different "workspace cleanup PR", but I think I'm fine with the default `workspace:^` version used by pnpm. I'm not even sure if the versions mean anything because it's found locally, but maybe using the default is better so that way `pnpm install <package from monorepo>` will have the correct workspace version (I'm pretty sure `workspace:*` should only exist if you manually type that into the `package.json`)

```suggestion
    "@libs/websoc-utils": "workspace:^",
```

apps/api/v1/rest/websoc/src/APILambdaClient.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(configuration: LambdaClientConfig = {}) {
+    this.client = new LambdaClient(configuration);
+    if (process.env.NODE_ENV === "development") {
+      import("@services/websoc-proxy").then((x) => (this.service = x.handler));
+    } else {
+      this.service = undefined;
+    }
+  }
+
+  private async invoke(body: Record<string, unknown>) {
```
ap0nia, line 24: Extending the other comment in this file, since this method is async, could we add an `initialized` checker. For example:
```ts
class APILambdaClient {
  initialized = false

  async ensureInitialized() {
    if (this.initialized) {
      return
    }
    await ...
  }

  async invoke() {
    await this.ensureInitialized()
    ...
  }
}
```
This is technically more overhead per invocation, but I think it's the most "correct" approach when dealing with these async shenanigans.

apps/api/v1/rest/websoc/src/lib.ts
```diff
@@ -53,101 +14,6 @@ export const fulfilled = <T>(value: PromiseSettledResult<T>): value is PromiseFu
  */
 export const notNull = <T>(x: T): x is NonNullable<T> => x != null;
```
ap0nia, line 15: Maybe a TODO would be to move this to `@libs/utils`, a personal package with miscellaneous, simple utils.

apps/docs/docs/developers-guide/rest-api/reference/websoc.md
```diff
@@ -127,7 +127,14 @@ curl "https://api-next.peterportal.org/v1/rest/websoc?year=2023&quarter=Spring&s
...
+
+<details>
+<summary>Section with no final exam</summary>
+
+```bash
+curl "https://api-next.peterportal.org/v1/rest/websoc?year=2023&quarter=Spring&sectionCodes=34271"
+```
 
 ```json
 {
```
ap0nia, line 140: Is there any way to import this JSON data from an external file, preferably one that's typed? It seems a bit scary that this data can become de-synced from what the API does since it's just plaintext markdown.

packages/peterportal-api-next-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "peterportal-api-next-types",
-  "version": "1.0.0-rc.2",
+  "version": "1.0.0-rc.2.68.2",
```
ap0nia, line 3: I don't think changeset uses particularly special versioning schemes, like `rc`, `alpha`, etc.
I don't think this is a big deal, just a heads up.

tools/grades-updater/src/sanitize-data.ts
```diff
@@ -3,10 +3,11 @@ import { EOL } from "os";
 import { basename, resolve } from "path";
```
ap0nia, line 3: Maybe we could flip an ESLint rule for this, but I personally like to specify `node` whenever possible.

```suggestion
import { basename, resolve } from "node:path";
```

tools/grades-updater/src/sanitize-data.ts
```diff
@@ -3,10 +3,11 @@ import { EOL } from "os";
...
 
 import { callWebSocAPI } from "@libs/websoc-api-next";
+import type { WebsocAPIResponse, WebsocSection } from "@libs/websoc-api-next";
```
ap0nia, line 6: Oh, is there a discrepancy between the monorepo's `websoc-api-next` and the currently published `peterportal-api-types-next` ?

services/websoc-scraper-v2/index.ts
```diff
@@ -11,8 +10,9 @@ import type {
   WebsocSchool,
   WebsocSection,
   WebsocSectionMeeting,
-} from "peterportal-api-next-types";
-import { geCodes, sectionTypes } from "peterportal-api-next-types";
+} from "@libs/websoc-api-next";
+import { callWebSocAPI, getDepts, getTerms } from "@libs/websoc-api-next";
+import { geCodes, sectionTypes } from "@libs/websoc-api-next";
```
ap0nia, line 15:
```suggestion
import { callWebSocAPI, getDepts, getTerms, geCodes, sectionTypes } from "@libs/websoc-api-next";
```

ap0nia (approved):
Looks good! I wrote down some stuff that we can consider/backlog for the project, but they're not significant blockers for this PR

tools/grades-updater/src/sanitize-data.ts
ecxyzzy: Yeah, they've diverged because of this feature, since the normalization happens in the route handler and not the module we use to fetch data from WebSoc.

apps/api/v1/rest/websoc/package.json
ecxyzzy: Tracked in #89.

apps/api/v1/rest/websoc/src/lib.ts
ecxyzzy: Tracked in #90.

apps/docs/docs/developers-guide/rest-api/reference/websoc.md
ecxyzzy: Generally when looking for examples, I try to find data that (if applicable) is in the past, so it's highly unlikely to change in the future. Also it serves as a sanity check whenever I modify the endpoints and need to redocument things, but that's just me.

packages/peterportal-api-next-types/package.json
ecxyzzy: This needs to be changed back to rc.2 before merging anyway.

ecxyzzy:
Tested locally by entering the incorrect password for the database URL, and the requests definitely took longer so I think that's a pretty clear indicator of getting the data from WebSoc.

PR #69, opened 2023-08-11
feat: :sparkles: devmoji

ap0nia:
# Summary
## Features
- Interops well with `commitlint` for locally verifying commits, and `conventional pr` for verifying PRs during CI/CD.
- Automatically adds an [shortcode style emoji](https://github.com/folke/devmoji#devmoji-emojify) to commit messages after being verified by `commitlint` during the `commit-msg` hook.
- Automatically adds an [shortcode style emoji](https://github.com/folke/devmoji#devmoji-emojify) to PR titles after being verified by `conventional PR` during CI/CD.

## Under the Hood (conventional PR, CI/CD)
After `conventional PR` verifies the title, `npx devmoji ...` is executed on the current PR title and [set as an output parameter](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter)

This output is passed as the `title` parameter in a PATCH request to the `update PR` GitHub API endpoint, which updates the PR title.
 
## Previous Pitch
Extremely important breaking change!

- [devmoji](https://github.com/folke/devmoji) automatically adds emojis to messages that are properly formatted according to commitlint.
- It's added as a `prepare-commit-msg` hook that will run after the `commitlint` checker 
  - Although we could use the `devmoji --lint` option and remove commitlint altogether, having both allows us to configure them separately
  - devmoji is configured with a separate config file: https://github.com/folke/devmoji#gear-configuration

# Issues
- It seems conventional PR gets upset at emojis in PR titles; should PR titles be allowed to have emojis?
- Oops, I probably should've marked this as `no-deploy` 😅

### DIFF
--- a/.github/workflows/conventional-pr.yml
+++ b/.github/workflows/conventional-pr.yml
@@ -1,9 +1,10 @@
 name: Lint PR title and message
 
+permissions:
+  pull-requests: write
+
 on:
   pull_request_target:
-
-  workflow_dispatch:
 
 jobs:
   lint-pr:
@@ -24,13 +25,14 @@ jobs:
       - name: Format PR Title
         id: devmoji
         run: |
-          echo TITLE=$(npx devmoji --format shortcode --text ${{ github.event.pull_request.title }}) >> "$GITHUB_OUTPUT"
+          echo TITLE=$(npx devmoji --format shortcode --text ${{ toJSON(github.event.pull_request.title) }}) >> "$GITHUB_OUTPUT"
 
       - uses: octokit/request-action@v2.x
         with:
           route: PATCH /repos/{owner}/{repo}/pulls/{pull_number}
+          owner: ${{ github.repository_owner }}
           repo: ${{ github.event.repository.name }}
           pull_number: ${{ github.event.number }}
-          title: ${{ steps.devmoji.outputs.TITLE }}
+          title: ${{ toJSON(steps.devmoji.outputs.TITLE) }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/conventional-pr.yml
+++ b/.github/workflows/conventional-pr.yml
@@ -45,7 +45,7 @@ jobs:
         run: pnpm install --dev --frozen-lockfile --ignore-scripts --workspace-root
 
       - name: Lint PR title
-        run: echo ${{ github.event.pull_request.title }} | npx --no -- commitlint
+        run: echo ${{ toJSON(github.event.pull_request.title) }} | npx --no -- commitlint
 
       - name: Format PR title
         id: devmoji
--- a/.github/workflows/conventional-pr.yml
+++ b/.github/workflows/conventional-pr.yml
@@ -10,18 +10,31 @@ jobs:
   lint-pr:
     runs-on: ubuntu-latest
     steps:
+      - name: Check out codebase
+        uses: actions/checkout@v3
+
       - name: Set up Node.js environment
         uses: actions/setup-node@v3
         with:
           node-version: 18
 
+      - name: Install pnpm
+        uses: pnpm/action-setup@v2
+        id: pnpm-install
+        with:
+          version: 8
+          run_install: false
+
+      - name: Install devDpendencies
+        run: pnpm install --dev --frozen-lockfile --ignore-scripts --workspace
+
       - name: Lint PR title
-        run: echo ${{ github.event.pull_request.title }} | npx commitlint
+        run: echo ${{ github.event.pull_request.title }} | npx --no -- commitlint
 
       - name: Format PR title
         id: devmoji
         run: |
-          echo TITLE=$(npx devmoji --format shortcode --text ${{ toJSON(github.event.pull_request.title) }}) >> "$GITHUB_OUTPUT"
+          echo TITLE=$(npx --no -- devmoji --format shortcode --text ${{ toJSON(github.event.pull_request.title) }}) >> "$GITHUB_OUTPUT"
 
       - name: Update PR title
         uses: octokit/request-action@v2.x
--- a/.github/workflows/conventional-pr.yml
+++ b/.github/workflows/conventional-pr.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   lint-and-format-pr:
-    runs-on: "ubuntu-latest"
+    runs-on: ubuntu-latest
     steps:
       - name: Check out codebase
         uses: actions/checkout@v3
--- a/.github/workflows/conventional-pr.yml
+++ b/.github/workflows/conventional-pr.yml
@@ -16,14 +16,14 @@ jobs:
           node-version: 18
 
       - name: Lint PR title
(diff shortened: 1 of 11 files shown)


ecxyzzy:
Don't really see the point for this, considering we squash everything before merging anyways and people probably won't bother adding the correct emoji to the PR title.

ap0nia:
The emojis are added automatically after you write the commit message; the default name for the PR would be the commit message that already has the emoji.

But right now, the PR linter doesn't even like emojis in the PR title lol.

ecxyzzy:
For branches with multiple commits, the name seems to default to some form of the branch name. So this would only work when someone opens a PR after having only made one commit. This is not something that should be relied upon.

ap0nia:
Ok, now...

- The `conventional-pr` workflow will automatically add a shortcode emoji to the PR if it's formatted per the [`conventional-commit`](https://www.conventionalcommits.org/en/v1.0.0/) standards. (To be clear, the `conventional-commit` standard is what `namchee/conventional-pr` is verifying against).

- The `commit-msg` hook will automatically add a shortcode emoji to the commit message if it's formatted correctly per `commitlint` standards.

The result:
- All squashed commits that appear on the main branch are formatted per the `conventional-commit` standard will have cute shortcode emojis and a list of squashed commits that are formatted per `commitlint` standards and also have cute shortcode emojis.
- All commits made to a development branch will be formatted per `commitlint` standards and have cute shortcode emojis
- All emojis are automatically added as long as the commit message or PR title conforms to their corresponding standard.

libs/websoc-api-next/package.json
```diff
@@ -12,7 +12,7 @@
     "test:watch": "vitest run --coverage --watch"
   },
   "dependencies": {
-    "@ap0nia/camaro": "6.2.4",
+    "@ap0nia/camaro": "6.2.5",
```
ap0nia, line 15: I was trying to unpublish my piscina package, but I misclicked and killed the camaro package instead, so I had to publish a new version lol

.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+npx --no -- lint-staged
```
ap0nia, line 4: `--no` indicates `no install`, i.e. don't install this dependency because it should be found in the current project. We install everything we need for git hooks as `devDependencies` in the project root, so I think it's ok to opt into this whenever possible.

The only issue I see is if we wanted to trigger the hook during CI/CD, but we didn't install `node_modules` or some obscure edge case like that.

.github/workflows/conventional-pr.yml
```diff
@@ -1,13 +1,38 @@
...
+
 jobs:
   lint-pr:
     runs-on: ubuntu-latest
     steps:
+      - name: Set up Node.js environment
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
+
```
ap0nia, line 17: Maybe this step could be cached somehow? I literally just wanna run `npx devmoji` lol

ecxyzzy:
Hmm yeah, I like this iteration much better.

Quick thought: Since we're running a Node script to generate the devmoji-ified PR name anyways, what if we got rid of the `conventional-pr` workflow and verified the PR name using `commitlint` since we already have it installed as a devDependency? That would also allow the PR name to have an emoji instead of a shortcode, since right now if I try to merge it, it shows the shortcode in the commit message instead.
image

ap0nia:
I think I prefer shortcode emojis tbh; the raw text message displays the emoji code, but the actual emoji is displayed when you view it from the GitHub UI. I think it's easier to stay consistent by using shortcode emojis (i.e. that's what I use when commenting and stuff) everywhere, and relying on the UI to render out the correct emoji.

ecxyzzy:
Yeah makes sense. Still think we should consider using `commitlint` though, since the `conventional-pr` action requires a Docker build, which is blazingly slow in my experience :^)

ap0nia:
Oh yeah, using commitlint just to bypass that seems valid lol. I don't get why it has to build anything, why not just put the binary up on AUR or something :^)

Feel free to modify the workflow to your liking to accommodate this!

ecxyzzy:
:rocket:
image

Let me know if you're cool with my changes, since all of yours look fine to me.

.github/workflows/conventional-pr.yml
```diff
@@ -1,13 +1,49 @@
-name: Lint PR title and message
+# yaml-language-server: $schema=https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/schemas/json/github-workflow.json
```
ap0nia, line 1: TIL a [YAML LSP](https://github.com/redhat-developer/yaml-language-server) exists that basically works like the JSON LSP. :rocket:

image

turbo.json
```diff
@@ -1,4 +1,5 @@
 {
+  "$schema": "https://turbo.build/schema.json",
```
ap0nia, line 2: This is for autocomplete with the JSON LSP; all JSON files can contain this field to ensure autocompletion, but a lot of editors automatically assign schemas to certain file names out of the box.

e.g. The most obvious one being `package.json` usually being associated with some schema like this: https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/package.json. I'm more aware of this because Vim has a feature :tm: where these associations don't exist out of the box :sunglasses: 

Most editors probably wouldn't include an association with `turbo.json` out of the box, which is why you'd specify `$schema` here instead of more common JSON files like `package.json` or `tsconfig.json`.

ap0nia:
Looks ready to ship 🚢

turbo.json
ecxyzzy: WebStorm supremacy :sunglasses:
image
(this screenshot was taken without the `$schema` field)

Jokes aside, I can see how this change would improve DX for people who use other editors. I'll add a task for it.

ecxyzzy:
> A small TODO could be to rename the file once it's on the main branch
Should be fine to just do it now.

.github/workflows/lint-pr.yml
```diff
@@ -0,0 +1,63 @@
...
+        id: pnpm-cache
+        shell: bash
+        run: echo "STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
+
+      - name: Setup pnpm cache
+        uses: actions/cache@v3
+        with:
+          path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
+          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
+          restore-keys: ${{ runner.os }}-pnpm-store-
```
ecxyzzy, line 39: image

New issue for adding pnpm cache where applicable? :eyes:

ap0nia: For my next ~~trick~~ PR, I was gonna clean up the workflow files; I was gonna add a `.github/actions/checkout-pnpm/action.yml` custom action that would checkout the repository and install Node, pnpm, pnpm cache, etc. that we could reuse on anything that requires setting that stuff up - i.e. TypeScript checks, the new PR linting strategy, etc.

PR #75, opened 2023-08-15
feat: :sparkles: implement arbitrary filtering for courses and instructors

### LINKED ISSUE #74: Arbitrary filtering for courses and instructors

ecxyzzy:
Closes #74.


apps/api/v1/rest/courses/src/lib.ts
```diff
@@ -52,3 +54,44 @@ export function normalizeCourse(course: PrismaCourse): Course {
...
 }
+
+/**
+ * Constructs a Prisma query for the given filter parameters.
+ * @param parsedQuery The query object parsed by Zod.
+ */
+export function constructPrismaQuery(parsedQuery: Query): Prisma.CourseWhereInput {
+  const AND: Prisma.CourseWhereInput[] = [];
+
+  if (parsedQuery.department) AND.push({ department: parsedQuery.department });
```
ap0nia, line 65: Personally, I just kinda decided that I like always using braces just so everything looks super consistent. Sometimes not including braces looks fine, but other times it doesn't and then I get confused what to look for lol. Not really a big deal, maybe something prettier can enforce? :shrug:

```suggestion
  if (parsedQuery.department) {
    AND.push({ department: parsedQuery.department })
  }
```

apps/api/v1/rest/courses/src/lib.ts
```diff
@@ -52,3 +54,44 @@ export function normalizeCourse(course: PrismaCourse): Course {
     terms: course.terms as unknown as string[],
   };
 }
+
+/**
+ * Constructs a Prisma query for the given filter parameters.
+ * @param parsedQuery The query object parsed by Zod.
+ */
+export function constructPrismaQuery(parsedQuery: Query): Prisma.CourseWhereInput {
+  const AND: Prisma.CourseWhereInput[] = [];
```
ap0nia, line 63: Using an array to store `AND` queries is a very clean solution for dynamic filtering 👍

apps/api/v1/rest/courses/src/schema.ts
```diff
@@ -1,5 +1,24 @@
+import { flattenStringsAndSplit } from "ant-stack/utils";
+import { divisionCodes, geCodes } from "peterportal-api-next-types";
 import { z } from "zod";
 
-export const QuerySchema = z.object({});
+export const QuerySchema = z.object({
```
ap0nia, line 5: One miscellaneous limitation I found is that JSDoc doesn't carry over from schema to inferred type (which makes sense, but just unfortunate).

e.g.
```ts
const schema = z.object({
 /**
  * Name of person.
  */
  name: z.string()
})

const Schema = z.infer<typeof schema>

/**
 * {
 *   (no JSDoc info)
 *   name: string
 * }
 */
```

The biggest issue with this arises if we use schemas to define a publicly consumable API; i.e. when users import an interface like `WebsocInput` that originated from a schema, they won't get any helpful JSDoc. Just something to keep in mind when we want to tackle the SDK; we might want to use TS interfaces/types with good JSDoc first, before using schemas for validating certain things.

packages/ant-stack/src/utils/schema-tools.ts
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Get unique, sorted array of strings.
+ * @param value String of comma-separated values or array of such strings.
+ */
+export const flattenStringsAndSplit = (value: TransformInput): TransformOutput =>
+  value
+    ? [
+        ...new Set(Array.isArray(value) ? value.flatMap((x) => x.split(",")) : value.split(",")),
+      ].sort()
```
ap0nia, line 21: I think I actually prefer `Array.from` syntax more than array spreading, just because the brackets can sometimes cause syntax confusion (i.e. expressions surrounded by brackets or parentheses immediately followed by more brackets/parentheses can be interpreted as a function call, which Prettier commonly addresses with a semicolon)

```suggestion
    ? Array.from(new Set(Array.isArray(value) ? value.flatMap((x) => x.split(",")) 
    : value.split(","))).sort()
```

packages/ant-stack/src/utils/schema-tools.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+export const flattenDayStringsAndSplit = (value: TransformInput): TransformOutput =>
+  value
+    ? [
+        ...new Set(
+          Array.isArray(value)
+            ? value.flatMap((x) => days.filter((y) => y.includes(x)))
+            : days.filter((x) => value.includes(x)),
+        ),
+      ]
```
ap0nia, line 36: Try `Array.from` here too

ap0nia (changes requested):
I think everything looks reasonable and well written :rocket:. Some minor styling concerns (that could be enforced by prettier?) I had were:

- using braces for ***all*** conditional statements (personal preference)
- using `Array.from` instead of `[...]` because of syntax confusion

braces anecdote: professor Klefstad said he preferred to exclude braces whenever possible to keep the syntax clean and easy to parse. idk, I just seem to incur mental overhead when figuring out if the braces are there or not lol :shrug:

apps/api/v1/rest/courses/src/lib.ts
ecxyzzy: Looks like there's an ESLint rule for that™: https://eslint.org/docs/latest/rules/curly#all

I can add it along with all of the other rules that we wanted to add before, since some of them require additional experimentation and are otherwise out of scope for this PR.

apps/api/v1/rest/courses/src/schema.ts
ecxyzzy: Yeah this is definitely something we'll have to figure out down the line. I don't really intend for the SDK to use the schema; this is purely for parsing unverified input directly from a REST client.

ap0nia (approved):
Looks good to me, good work! :fire:

ap0nia (approved):
Documentation looks good :+1:

An interesting plugin for docusaurus you could check out is [typedoc](https://www.npmjs.com/package/docusaurus-plugin-typedoc); it can parse source code and then generate markdown content with the JSDoc. The auto-generated content isn't the nicest looking stuff tbh, but it's always up to date with no effort :shrug:

Example of auto-generated docs: https://trpc.io/docs/typedoc/client/

PR #85, opened 2023-08-23
fix: :bug: make sure average GPA field is not NaN

### LINKED ISSUE #83: Cannot return null on non-nullable field
I think this should work. The structure worked for other queries but not this one, and it's not clear to me why that's the case. Everything works as long as the instructor's name isn't provided, which is what AntAlmanac has been using until now, hence why this was not discovered.

```graphql
{ 
        aggregateGrades(department: "I&C SCI", courseNumber: "45C", instructor: "SHINDLER, M.") {
            gradeDistribution {
                gradeACount
                gradeBCount
                gradeCCount
                gradeDCount
                gradeFCount
                gradePCount
                gradeNPCount
                averageGPA
            }
        },
    }
```

I'm marking this as high because it breaks the feature being introduced to Antalmanac (icssc/AntAlmanac#671)

ecxyzzy:
Closes #83.


MinhxNguyen7 (approved):
LGTM!

PR #86, opened 2023-08-26
fix: 🐛 use the custom devmoji config

ap0nia:
# Summary
It turns out `devmoji` specifically reads from `devmoji.config.js` not from any variants like `devmoji.config.cjs` by default. This incorrect behavior can be verified by adding adding a `revert` entry in the config, and creating a Git commit message with "revert" as the type -- it doesn't pick up the new entry.

- Added a `revert` type
- Made the `types` config property derived from the emojis supported.
- Use real emojis instead of shortcode ones


ecxyzzy (approved):
LGTM :rocket:

PR #88, opened 2023-08-28
feat: ✨ add filtering by GE and additional aggregation operations

### LINKED ISSUE #87: Grades endpoint: operation for aggregate grades data by same course/instructor pair
See: icssc/AntAlmanac#671

### LINKED ISSUE #94: Grades endpoints: Add filtering by GE

ecxyzzy:
## Summary
Add filtering by GE category for all endpoints, as well as the aggregate grouped operation. The section listing for raw and aggregate grades also include the section's GE categor(y/ies).

The aggregate grouped operation returns a list containing the aggregated grade distribution for each (course, instructor) pair.

## Notes

This PR used to implement only the latter feature (hence the branch name), but the first change was later determined to be in scope for this PR and so was built upon this PR.

## Issues

Closes #87; closes #94.


MinhxNguyen7 (changes requested):
I haven't looked at the code yet, so I'm only going to comment on functionality.

This is good for when the user is looking at a department or a particular course. However, a (arguably more) common use case is selecting a GE with the highest GPA. This feature does not support that use case more than the current one does. 

I'm not sure what you can do about that. Perhaps implementing selection by GE is as easy as by department; I don't know how your database is structured.

apps/api/v1/rest/grades/src/lib.ts
```diff
@@ -96,3 +205,61 @@ export function aggregateGrades(grades: GradesRaw): GradesAggregate {
...
+    }
+  }
+  return Array.from(courses)
+    .map(([k, v]) => ({
+      ...(Object.fromEntries(
+        (JSON.parse(k) as string[]).map((x, i) => [headerKeys[i], x]),
+      ) as AggregateGradeByCourseHeader),
+      ...aggregateGrades(v).gradeDistribution,
+    }))
+    .sort((a, b) => lexOrd(a.department, b.department) || lexOrd(a.courseNumber, b.courseNumber));
```
MinhxNguyen7, line 232: For out purposes, we don't need the response to be sorted, but idk if this is just standard.

MinhxNguyen7 (changes requested):
When querying aggregateByCourse, there's an option to put instructor in the metadata, but that results in an unhelpful error: "`Cannot return null for non-nullable field AggregateGradeByOffering.instructor`". It doesn't make sense to aggregate by instructor, but that is an option, so it should either be removed or at least given a more helpful message.

Everything else seems fine. I'll incorporate the aggregateByCourses endpoint into the AntAlamanac PR for more testing.

I can't speak much to the goodness of the code. It looks alright I guess.

MinhxNguyen7 (approved):
LGTM!

PR #96, opened 2023-09-07
feat: ✨ fourmilier devops

ap0nia:
# Summary
Devops solved

# Next Steps
No more ~~cooking~~ devops please :pray:


apps/api-fourmilier/bronya.config.ts
```diff
@@ -0,0 +1,137 @@
+import dotenv from "dotenv";
+
+dotenv.config({ path: "../../.env" });
```
ecxyzzy, line 3: Should probably just use `dotenv/config` here since GitHub Actions just puts all the secrets in the runner environment.

apps/api-fourmilier/bronya.config.ts
```diff
@@ -0,0 +1,137 @@
...
+          queryEngines
+            .filter((x) => x !== "libquery_engine-linux-arm64-openssl-1.0.x.so.node")
+            .forEach((queryEngineFile) => {
+              fs.rmSync(path.join(directory, queryEngineFile));
+            });
+        },
+      },
+      environment: {
+        DATABASE_URL: process.env["DATABASE_URL"] ?? "",
+        SHADOW_DATABASE_URL: process.env["SHADOW_DATABASE_URL"] ?? "",
```
ecxyzzy, line 73: The shadow database URL isn't actually necessary.

apps/api-fourmilier/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "homepage": "https://github.com/icssc/peterportal-api-next",
+  "bugs": {
+    "url": "https://github.com/icssc/peterportal-api-next/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/icssc/peterportal-api-next",
+    "directory": "apps/api"
+  },
+  "funding": "https://github.com/icssc/peterportal-api-next",
```
ecxyzzy, line 16:
```suggestion
```

apps/api-fourmilier/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "aws-cdk": "^2.93.0",
+    "dotenv": "^16.3.1",
+    "dotenv-cli": "^7.3.0",
+    "tsx": "^3.12.7"
+  },
+  "packageManager": "pnpm@8.6.12",
+  "engines": {
+    "node": ">=18",
+    "pnpm": "^8.0.0"
+  }
```
ecxyzzy, line 52: Dependencies and engines should all be pinned.

apps/api-fourmilier/src/v1/rest/websoc/APILambdaClient.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.client = new LambdaClient(configuration);
+  }
+
+  static async new(configuration: LambdaClientConfig = {}) {
+    const client = new APILambdaClient(configuration);
+
+    await client.$connect();
+
+    return client;
+  }
```
ecxyzzy, line 38: Is there actually any reason to keep this if the constructor is public and we have the `$connect` method? Or is the constructor being public not actually intentional?

apps/api/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "dependencies": {
+    "peterportal-api-next-types": "workspace:1.0.0-rc.3"
+  },
+  "devDependencies": {
+    "@types/aws-lambda": "^8.10.119"
```
ecxyzzy, line 6: Pin pls

libs/lambda/package.json
```diff
@@ -0,0 +1,16 @@
...
+  "license": "MIT",
+  "type": "module",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "dependencies": {
+    "winston": "3.10.0"
+  },
+  "devDependencies": {
+    "@peterportal-api/types": "workspace:^",
+    "@types/aws-lambda": "^8.10.119"
```
ecxyzzy, line 14: Also pin pls

apps/api/package.json
ap0nia: Would you want me to just go through and pin everything real quick?

What if we just force ran renovate on the monorepo again :^)

apps/api-fourmilier/src/v1/rest/websoc/APILambdaClient.ts
ap0nia: Oh, I temporarily used `moduleResolution: "bundler"` in the tsconfig because the exports from my package weren't being resolved with `moduleResolution: "Node"`. This setting didn't allow top-level async, so I wrote a `$connect` method to initialize it in the async handler. I figured out what was wrong, and now we're using `moduleResolution: "Node"` again, which does allow me to just do `await APILambdaClient.new()`

I could remove this if we want to continue initializing with top-level await.

libs/lambda/src/compress.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (body.length <= MIN_COMPRESSION_SIZE) {
+    return {
+      body: gzipSync(body).toString("base64"),
+      method: "gzip",
+    };
+  }
+
+  if (!acceptEncoding) {
+    return { body };
+  }
```
ecxyzzy, line 46: I guess I should have documented it, but per the [relevant section](https://www.rfc-editor.org/rfc/rfc9110#section-12.5.3-10.1) in the HTTP Semantics RFC, if `Accept-Encoding` is not provided, then any encoding should be acceptable. So we should use gzip if it's greater than the compression threshold. We should only return the payload in plaintext if it's below the compression threshold, or if `Accept-Encoding` is the empty string.

libs/lambda/src/compress.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export function compress(body: string, acceptEncoding?: string): CompressionResult {
+  // Default to using gzip if the body size is greater than the threshold.
+  if (body.length <= MIN_COMPRESSION_SIZE) {
+    return {
+      body: gzipSync(body).toString("base64"),
+      method: "gzip",
+    };
+  }
```
ecxyzzy, line 42: Logic error here maybe? Isn't this less than the minimum compression size?

apps/api-fourmilier/bronya.config.ts
ap0nia: Oh, if this is intended to be run only during devops, then I can remove dotenv from the script altogether. I was deploying the canary version manually, so I needed to read from my local file system's env file.

apps/api-fourmilier/bronya.config.ts
ap0nia: Should I completely remove it then?

libs/lambda/src/compress.ts
ap0nia: Oh, I copy/pasted the wrong line of code lol.

package.json
```diff
@@ -28,8 +28,11 @@
     "@typescript-eslint/eslint-plugin": "6.5.0",
     "@typescript-eslint/parser": "6.5.0",
     "ant-stack": "^0.1.0",
+    "arktype": "1.0.16-alpha",
+    "aws-cdk-lib": "^2.93.0",
     "cz-conventional-changelog": "3.3.0",
     "devmoji": "2.3.0",
+    "dotenv": "^16.3.1",
```
ecxyzzy, line 35: Also also pin pls

packages/types/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "homepage": "https://github.com/icssc/peterportal-api-next",
+  "bugs": {
+    "url": "https://github.com/icssc/peterportal-api-next/issues"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/icssc/peterportal-api-next",
+    "directory": "packages/types"
+  },
+  "funding": "https://github.com/icssc/peterportal-api-next",
```
ecxyzzy, line 16:
```suggestion
```

packages/websoc-fuzzy-search/package.json
```diff
@@ -18,7 +18,8 @@
   },
   "dependencies": {
     "base64-arraybuffer": "1.0.2",
-    "pako": "2.1.0"
+    "pako": "2.1.0",
+    "peterportal-api-next-types": "workspace:1.0.0-rc.3"
```
ecxyzzy, line 22: Wait why does fuzzy search depend on types lol

ap0nia: This file apparently uses it https://github.com/icssc/peterportal-api-next/blob/fourmilier/packages/websoc-fuzzy-search/setup.ts

tsconfig.json
```diff
@@ -3,8 +3,8 @@
     "baseUrl": ".",
     "strict": true,
     "outDir": "dist",
-    "target": "ES2020",
-    "module": "ES2020",
+    "target": "ESNext",
+    "module": "ESNext",
```
ecxyzzy, line 7: I was actually thinking of using [@tsconfig/node18](https://www.npmjs.com/package/@tsconfig/node18) and keeping `"module": "es2022"` and `"moduleResolution": "node"`. Thoughts?

tsconfig.json
```diff
@@ -13,15 +13,6 @@
...
     "moduleResolution": "Node",
-    "paths": {
-      "@libs/*": ["libs/*"],
-      "ant-stack": [
-        "packages/ant-stack/src",
-        "packages/ant-stack/src/lambda-core",
-        "packages/ant-stack/src/utils"
-      ],
-      "peterportal-api-next-types": ["packages/peterportal-api-next-types"]
-    },
```
ecxyzzy, line 24: Why did this get killed?

ap0nia: Because if TypeScript is setup and working properly, then this shouldn't be needed. The packages should be composed correctly (this entails properly detailing the "typesVersions" property in the package's package.json if it has exports from sub-directories)

"typesVersions" is read when using moduleResolution: Node; otherwise bundler, NodeNext, and Node16 will read the "exports" field.

tsconfig.json
ap0nia: I don't really care as long as top-level await is supported. Just having it available generally makes life easier :nerd_face:

.npmrc
```diff
@@ -1,4 +1 @@
-enable-pre-post-scripts = true
-shamefully-hoist = true
 engine-strict = true
-link-workspace-packages = true
```
ecxyzzy, line 4: Yeah this looks fine for all of the services; we were probably doing some off-label usage beforehand.

ecxyzzy (changes requested):
Nice start, thank you so much for working on all of this. :pray: Some changes are still necessary though.

Also any reason you didn't just nuke the original API and add this as a drop-in? It should be fine as long as we don't merge before everything is complete.

ap0nia:
> Also any reason you didn't just nuke the original API and add this as a drop-in?
Not really, I was thinking of just keeping it around as a reference or something. I can remove it now if you'd like.

apps/api-fourmilier/bronya.config.ts
ecxyzzy: I'm pretty sure the workspace root's `dev` script already injects development credentials. So yeah, I think this is unnecessary, as long as the API has a `dev` script.

apps/api-fourmilier/bronya.config.ts
ecxyzzy: Yeah please do :pray:

apps/api/package.json
ecxyzzy: I think that's less hassle than rerunning renovate, so please do :pray:

packages/websoc-fuzzy-search/package.json
ecxyzzy: That's a setup script and not actually included in the final bundle so it should probably go under devDeps.

ecxyzzy:
Yeah I think that would make sense, and then we could adapt the staging stuff accordingly as well.

apps/api-fourmilier/src/v1/rest/websoc/APILambdaClient.ts
ecxyzzy: Yeah I think the static async factory pattern makes the most sense since we have top-level await.

tsconfig.json
ecxyzzy: Will include it in my PR against this branch.

apps/api/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "dev": "bunny dev-api"
+  },
+  "dependencies": {
+    "@aws-sdk/client-lambda": "3.398.0",
+    "@libs/db": "workspace:^",
+    "@libs/lambda": "workspace:^",
+    "@libs/websoc-utils": "workspace:^",
+    "@peterportal-api/types": "workspace:^",
+    "@services/websoc-proxy": "workspace:^",
+    "@uc-irvine/api": "workspace:^",
```
ecxyzzy, line 30: Already discussed privately but this should probably go under `@libs/ucirvine-api` or similar.

apps/api/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@apps/api",
+  "version": "0.0.0",
+  "private": true,
+  "description": "core lambda api for peterportal api",
```
ecxyzzy, line 5:
```suggestion
  "description": "The workspace for PeterPortal API",
```

apps/docs/package.json
```diff
@@ -38,8 +38,5 @@
     "@tsconfig/docusaurus": "2.0.0",
     "@types/node": "18.17.12",
     "typescript": "5.2.2"
-  },
-  "engines": {
-    "node": "18"
```
ecxyzzy, line 43: Is the engines field not that relevant for a workspace package.json?

ecxyzzy (changes requested):
Cooking :fire:

Some slight changes needed still; feel free to address them at your leisure. Will start on a PR for the graphql stuff soon™.

apps/docs/package.json
ap0nia: tbh, I have no idea lol. My impression is that it's only significant to the root workspace, since that's where installations originate (i.e. the root node_modules). 

Weighing the pros/cons of maintaining the "engines" field for all private projects: I think it isn't really worth it, especially since everything is pinned. i.e. If we ever change one engine, we should update everybody else's, despite them not actually doing anything. I guess this is only an issue if renovate doesn't take care of it automatically. I think I'm ok with maintaining fully fledged package.jsons for officially released packages, and then keeping them simpler for things that are internal.

ecxyzzy (approved):
Ship ittt!! :rocket: :rocket: :rocket: 

Once again thank you so much for working on this. DX is going to be fire going forward!

PR #97, opened 2023-09-08
feat: ✨ migrate graphql endpoint

ecxyzzy:


apps/api/bronya.config.ts
```diff
@@ -9,15 +9,13 @@ import { createApiCliPlugins } from "@bronya.js/api-construct/plugins/cli";
...
 const js = `\
-import * as path from 'path';
-import { fileURLToPath } from 'url';
-import { createRequire as topLevelCreateRequire } from 'module';
-const require = topLevelCreateRequire(import.meta.url);
-const __filename = fileURLToPath(import.meta.url);
-const __dirname = path.dirname(__filename);
+const require = (await import("node:module")).createRequire(import.meta.url);
+const __filename = (await import("node:url")).fileURLToPath(import.meta.url);
+const __dirname = (await import("node:path")).dirname(__filename);
```
ap0nia, line 18: Not too important

I think my preferred strategy was this, since it doesn't deal with top-level await, which sometimes doesn't work:

```ts
import topLevelModule from 'node:module'
import topLevelUrl from 'node:url'
import topLevelPath from 'node:path'

const require = topLevelModule.createRequire(import.meta.url);
const __filename = topLevelUrl.fileURLToPath(import.meta.url);
const __dirname = topLevelPath.dirname(__filename);
```

There's also strategies with dynamically creating some random hash as a variable and doing string substitution to make naming conflicts even less likely?

Something interesting Vite does with regards to __filename and __dirname, is to define [custom strings as variable names here](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/config.ts#L1012), and then replace them with a [custom esbuild plugin when loading the file](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/config.ts#L1137)

Nothing here actually has to be implemented or anything, it's was just kind of interesting how developers have monkey-patched their own solutions lol

apps/api/bronya.config.ts
```diff
@@ -80,19 +78,33 @@ class MyStack extends Stack {
...
         format: "esm",
         platform: "node",
         bundle: true,
+        minify: true,
         banner: { js },
         outExtension: { ".js": ".mjs" },
         plugins: [
           {
-            name: "copy",
+            name: "copy-graphql-schema",
```
ap0nia, line 91: Is this required during dev? I think the most correct thing to do is handle this above, in the `constructs.lambdaUpload`. The main motivation is to reduce the amount of work to build a particular route; i.e. the copy-prisma plugin should be safely migrated to `constructs.lambdaUpload`, but I haven't gotten around to it.

It's not too big of a deal, because I still need to do some testing and stuff to make sure everything is stable regardless of where we choose to manipulate files.

apps/api/src/routes/v1/graphql/+endpoint.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const req: HTTPGraphQLRequest = {
+    body,
+    headers: Object.entries(eventHeaders).reduce(
+      (m, [k, v]) => m.set(k, Array.isArray(v) ? v.join(", ") : v ?? ""),
+      new HeaderMap(),
+    ),
+    method,
+    search: parse(event.path ?? "").search ?? "",
+  };
```
ap0nia, line 45: Maybe this strategy could be worth checking out to make the object lighter? The idea is to lazily compute and cache the headers; i.e. only if they're actually accessed at least once during the request.

```suggestion
  let headers: Record<string, string>
  const req: HTTPGraphQLRequest = {
    body,
    get headers() {
      headers ??= Object.entries(eventHeaders).reduce(
        (m, [k, v]) => m.set(k, Array.isArray(v) ? v.join(", ") : v ?? ""),
        new HeaderMap(),
      )
      return headers
    },
    method,
    search: parse(event.path ?? "").search ?? "",
  };
```

apps/api/src/routes/v1/graphql/+endpoint.ts
```diff
@@ -0,0 +1,80 @@
...
+    httpGraphQLRequest: req,
+    context: async () => ({}),
+  });
+
+  const statusCode = res.status ?? 200;
+  const headers: APIGatewayProxyResult["headers"] = {
+    "Access-Control-Allow-Headers": "Apollo-Require-Preflight, Content-Type",
+    "Access-Control-Allow-Origin": "*",
+    "Access-Control-Allow-Methods": "GET, POST, OPTIONS",
+  };
```
ap0nia, line 56: We can initialize this at the top level of the module, right? So we don't have to rebuild the object on every request.

ap0nia (approved):
Thanks for GraphQL! :pray:

The only actual change that could be implemented trivially is the lazily computed headers. The other comments are just considerations.

apps/api/bronya.config.ts
ecxyzzy: Yeah this would probably require the least amount of shenanigans. I just figured top-level await is cleaner since our env supports it now.

apps/api/bronya.config.ts
ecxyzzy: Yeah, this is required during dev. If there's a way to have this plugin only be active for the graphql route (I tried exporting an override but it didn't seem to want to work), that'd definitely be better.

PR #99, opened 2023-09-10
feat: ✨ restore deployment feature parity

ecxyzzy:
Now we're cooking :fire:

Side note: While not entirely within scope, I've taken the liberty of migrating to [@tsconfig/node18](https://www.npmjs.com/package/@tsconfig/node18) and removing some entries from the root tsconfig's compilerOptions that were already enforced by ESLint anyways.

### DIFF
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -95,7 +95,7 @@ class ApiStack extends Stack {
         format: "esm",
         platform: "node",
         bundle: true,
-        minify: true,
+        // minify: true,
         banner: { js },
         outExtension: { ".js": ".mjs" },
         plugins: [
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -160,6 +160,9 @@ class ApiStack extends Stack {
 }
 
 function getStage() {
+  if (!process.env.NODE_ENV) {
+    throw new Error("NODE_ENV not set.");
+  }
   switch (process.env.NODE_ENV) {
     case "production":
       return "prod";
@@ -181,10 +184,6 @@ export async function main() {
   const id = "peterportal-api-next";
 
   const app = new App();
-
-  if (!process.env.NODE_ENV) {
-    throw new Error("NODE_ENV not set.");
-  }
 
   const stage = getStage();
 
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -23,7 +23,7 @@ const js = `
 
   const require = topLevelModule.createRequire(import.meta.url);
   const __filename = topLevelUrl.fileURLToPath(import.meta.url);
-  const __dirname = topLeveldirname(__filename);
+  const __dirname = topLevelPath.dirname(__filename);
 `;
 
 const projectRoot = process.cwd();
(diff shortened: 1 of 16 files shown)


apps/api/bronya.config.ts
```diff
@@ -141,25 +160,99 @@ class MyStack extends Stack {
...
+      stage = `staging-${process.env.PR_NUM}`;
+      break;
+    case "development":
+      stage = "dev";
+      break;
+    default:
+      throw new Error(
+        "Invalid NODE_ENV specified. Valid values are 'production', 'staging', and 'development'.",
+      );
+  }
```
ap0nia, line 190: Can we encapsulate this as a function defined at the top level of the module (or maybe in `@libs/common` :eyes:)

```suggestion
// at the top-level
function getStage() {
  switch (process.env.NODE_ENV) {
    case "production":
      return "prod";
    case "staging":
      if (!process.env.PR_NUM) {
        throw new Error("NODE_ENV was set to staging, but a PR number was not provided.");
      }
      return `staging-${process.env.PR_NUM}`;
    case "development":
      return "dev";
    default:
      throw new Error(
        "Invalid NODE_ENV specified. Valid values are 'production', 'staging', and 'development'.",
      );
  }
}

const stage = getStage()

```

apps/api/bronya.config.ts
```diff
@@ -141,25 +160,99 @@ class MyStack extends Stack {
...
+      }),
+    );
+
+    result.api.methods.forEach((x) => {
+      try {
+        x.resource.addMethod("OPTIONS", optionsIntegration);
+      } catch {
+        // no-op
+      }
+    });
```
ap0nia, line 250: Nit
```suggestion
    result.api.methods.forEach((apiMethod) => {
      try {
        apiMethod.resource.addMethod("OPTIONS", optionsIntegration);
      } catch {
        // no-op
      }
    });
```

apps/api/bronya.config.ts
```diff
@@ -141,25 +160,99 @@ class MyStack extends Stack {
...
   }
 
   return app;
 }
 
 if (isCdk()) {
-  main();
+  main().then();
```
ap0nia, line 257: Would this work alright too?

```suggestion
  await main()
```

tsconfig.json
```diff
@@ -1,23 +1,16 @@
 {
+  "extends": "@tsconfig/node18/tsconfig.json",
```
ap0nia, line 2: What's the benefit of this? I've always written my tsconfigs from scratch, so maybe I've just been drinking the koolaid

ap0nia (approved):
Thanks, these changes look solid!

apps/api/bronya.config.ts
ecxyzzy: Yeah that's fine. I keep forgetting we have top level await now lol

tsconfig.json
ecxyzzy: Sensible defaults and fewer things for us to maintain? :shrug: I know Docusaurus and SvelteKit ship with their own tsconfigs at least, so I suppose there is some merit to it, but I guess that sounds koolaid-y.

ecxyzzy:
Sike again looks like cors is broken :upside_down_face:

PR #101, opened 2023-10-15
refactor: ♻️ quick config fix

ap0nia:
# Summary
- I have my env at the project root, so I'd like `dotenv` to just load that one for me when I happen to be in the `apps/api` directory
- Ideally, the built routes in the temporary `.bronya` directory don't actually need to include prisma because the dev server can find prisma just fine since it defaults to `node_modules` which exists locally. This speeds up the dev build because we don't have to manipulate any files before starting up. Also, I encountered some issues with "binary for ARM64" not found, which seemed to be mitigated when no copying takes place (because `node_modules` is a more reliable place to find prisma)

# Followup
- I think it would definitely be nice if the config file supported top-level await. Since I got tired of maintaining my custom JIT compiler, I switched back to JITI -- which actually happens [working on this feature](https://github.com/unjs/jiti/issues/72#issuecomment-1713502405)  :eyes:

### DIFF
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -16,6 +16,24 @@ import { App, Stack, Duration } from "aws-cdk-lib/core";
 import { config } from "dotenv";
 import type { BuildOptions } from "esbuild";
 
+/**
+ * Whether or not we're executing in CDK.
+ *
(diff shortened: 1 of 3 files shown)


ecxyzzy (approved):
:rocket:

PR #103, opened 2023-10-18
fix: 🐛 actually unbreak graphql (WIP)

ecxyzzy:


ecxyzzy:
I'm so confused because things are working again.

PR #107, opened 2023-10-18
feat: ✨ create handler helper function

ap0nia:
# Summary
- Add a `createHandler` function to the internal `@libs/lambda` package for API developers to use when creating new endpoints. 
- The function will check whether or not it's a warming request prior to executing the handler itself.

# Technical Details
- This function accepts a callback with the typical `event` and `context` parameters from API Gateway, along with a utility `res` object (similar to the one in Express.js) to generate a proper response object.
- The CDK framework doesn't handle warming requests out of the box, but we add them to the CDK after the relevant API Gateway resources are allocated. Then we handle the warming requests with a custom, internal handler wrapper.

### DIFF
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -86,7 +86,7 @@ export const esbuildOptions: BuildOptions = {
   format: "esm",
   platform: "node",
   bundle: true,
-  minify: true,
+  // minify: true,
   banner: { js },
 
   /**
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -4,7 +4,7 @@ import { join, resolve } from "node:path";
 import { Api, type ApiConstructProps } from "@bronya.js/api-construct";
 import { createApiCliPlugins } from "@bronya.js/api-construct/plugins/cli";
 import { isCdk } from "@bronya.js/core";
-import { logger } from "@libs/lambda";
+import { logger, warmingRequestBody } from "@libs/lambda";
 import { LambdaIntegration, ResponseType } from "aws-cdk-lib/aws-apigateway";
 import { Certificate } from "aws-cdk-lib/aws-certificatemanager";
 import { RuleTargetInput, Rule, Schedule } from "aws-cdk-lib/aws-events";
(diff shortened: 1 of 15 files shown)


apps/api/src/routes/v1/graphql/+config.ts
```diff
@@ -27,4 +27,8 @@ export const overrides: ApiPropsOverride = {
       },
     ],
   },
+  constructs: {
+    functionPlugin: constructs.functionPlugin,
+    restApiProps: constructs.restApiProps,
+  },
```
ap0nia, line 33: TODO: Documentation for the CDK framework (eventually :tm:)

Since override configs are shallow merged, the `constructs` property will remain the same as the root one unless defined again. The root one defines an uploading procedure that adds prisma, which we can safely omit for the GraphQL route to reduce the Lambda size.

apps/api/bronya.config.ts
```diff
@@ -79,21 +79,14 @@ const prismaSchema = resolve(libsDbDirectory, "prisma", prismaSchemaFile);
...
-
 /**
  * Shared ESBuild options.
  */
 export const esbuildOptions: BuildOptions = {
   format: "esm",
   platform: "node",
   bundle: true,
-  minify: true,
+  // minify: true,
```
ap0nia, line 89: I left the code un-minified for ease of debugging in prod, but we can re-enable it.

libs/lambda/src/handler.ts
```diff
@@ -0,0 +1,79 @@
...
+) => void | Promise<APIGatewayProxyResult | void>;
+
+/**
+ * Creates a handler for API Gateway.
+ *
+ * Handles warming requests and provides utilities for formatting responses.
+ */
+export function createHandler(
+  handler: ExtendedApiGatewayHandler,
+  onWarm?: ExtendedApiGatewayHandler,
```
ap0nia, line 57: Basically, you can pass two different event handlers to this helper function -- one to handle a regular request, and another to handle warming requests. This is currently primarily used to initialize prisma.

ecxyzzy (approved):
lgtm, not really sure why it took me so long. great work :tada:

PR #117, opened 2023-12-10
perf: ⚡️ optimize courses/instructors by ID

ecxyzzy:


apps/api/bronya.config.ts
```diff
@@ -97,12 +107,41 @@ export const esbuildOptions: BuildOptions = {
...
+          path: args.path,
+          namespace,
+        }));
+        build.onLoad({ filter: /virtual:courses/, namespace }, async () => ({
+          contents: `export const courses = ${JSON.stringify(
+            Object.fromEntries(
+              (await prisma.course.findMany()).map(normalizeCourse).map((x) => [x.id, x]),
+            ),
+          )}`,
+        }));
```
ap0nia, line 129: Not that anybody would ever look at this devops code or care what it does, but I'd personally implement it something like this so it's easier for me to understand. Same idea applies to the virtual instructors file, if you think it matters 🤓 

```suggestion
        build.onLoad({ filter: /virtual:courses/, namespace }, async () => {
         const allCourses =  await prisma.course.findMany()
         
         const courses = courses
           .map(normalizeCourse)
           .reduce((course) => ({
             [course.id]: course
           }), {} as Record<string, Course>)

         return {
            contents: `export const courses = ${JSON.stringify(courses)}`,
          }
        });
```

apps/api/src/global.d.ts
```diff
@@ -0,0 +1,13 @@
...
+// treats this as a normal module rather than an ambient module, and so the module declarations
+// don't actually work when consumed by other TypeScript source files.
+
+declare module "virtual:courses" {
+  declare const courses: Record<string, import("@peterportal-api/types").Course>;
+}
+
+declare module "virtual:instructors" {
+  declare const instructors: Record<string, import("@peterportal-api/types").Instructor>;
+}
```
ap0nia, line 13: Honestly, it would be great if I could find any official resource that actually describes ambient declaration files adequately, but I can't lol. I think jsdoc formatted like this might be more logical?

 ```suggestion
/**
 * Ambient declaration file for defining "virtual" modules/files. 
 * The file contents are generated dynamically during build time by esbuild.
 * DO NOT add any imports/exports; that converts the file to a regular module and removes the global declarations.
 * @see (link to ambient declaration information that I don't have)
 */


/**
 * Virtual module for caching course information during build time.
 */
declare module "virtual:courses" {
  declare const courses: Record<string, import("@peterportal-api/types").Course>;
}

/**
 * Virtual module for caching instructor information during build time.
 */
declare module "virtual:instructors" {
  declare const instructors: Record<string, import("@peterportal-api/types").Instructor>;
}
```

File: apps/api/src/lib/utils.ts [hunk 3]
```diff
@@ -35,3 +38,55 @@ export const flattenDayStringsAndSplit = (value: TransformInput): TransformOutpu
...
+      case "GE-6":
+        return "GE VI: Language Other Than English";
+      case "GE-7":
+        return "GE VII: Multicultural Studies";
+      case "GE-8":
+        return "GE VIII: International/Global Issues";
+      // this branch should never happen
+      default:
+        throw new Error();
+    }
```
ap0nia, line 80: This is technically out of scope of the PR, but what do you think of something like this?

```suggestion
   // At the top of the file somewhere
   const GE_CATEGORIES = {
     'GE-1A': 'GE 1a: Lower Division Writing',
     'GE-1B': 'GE 1b: Upper Division Writing',
     // ...
   }
   
    const category = GE_CATEGORIES[x]
    
    if (category = null) {
      throw new Error()
    }
    
    return category
```

ap0nia (approved):
I think everything looks solid 💪 

I just left styling remarks.

apps/api/src/lib/utils.ts
ecxyzzy: Yeah this should definitely have been a lookup table to begin with. I am not really sure what was going on inside my head when I wrote this code :upside_down_face:

apps/api/bronya.config.ts
ecxyzzy: The reduction into an object doesn't seem to actually work. Plus I think having the callback just return and not have a body is nice, but that's just me.

PR #118, opened 2023-12-21
fix(api): 🐛 merge construct props and function props correctly

ap0nia:
# Summary
Re-order the config merging.

# Details
The root function props were overriding the route-specific ones, which is problematic for the websoc routes because they lost the `role` property. 

I've monkey-patched production for now by adding a permission policy to the IAM role for the `/v1/rest/websoc` route.
image


ecxyzzy (approved):
:ship:

PR #119, opened 2023-12-28
feat: ✨ add previews for courses/instructors

ecxyzzy:
TODO before merging: copy dev schema courses/instructors table to prod.


apps/api/src/lib/utils.ts
```diff
@@ -64,11 +71,14 @@ export function normalizeCourse(course: PrismaCourse): Course {
   return {
     ...course,
     courseLevel,
-    instructorHistory: course.instructorHistory as unknown as string[],
+    instructorHistory: course.instructorHistory,
+    instructors: course.instructors as unknown as InstructorPreview[],
     prerequisiteTree: course.prerequisiteTree as unknown as PrerequisiteTree,
```
ap0nia, line 76: It seems to work for me without the first assertion to unknown.

```suggestion
    instructors: course.instructors as InstructorPreview[],
    prerequisiteTree: course.prerequisiteTree as PrerequisiteTree,
```

apps/api/src/lib/utils.ts
```diff
@@ -64,11 +71,14 @@ export function normalizeCourse(course: PrismaCourse): Course {
...
-    prerequisiteList: course.prerequisiteList as unknown as string[],
-    prerequisiteFor: course.prerequisiteFor as unknown as string[],
+    prerequisiteList: course.prerequisiteList,
+    prerequisiteFor: course.prerequisiteFor,
+    prerequisites: course.prerequisites as unknown as CoursePreview[],
+    dependencies: course.dependencies as unknown as CoursePreview[],
```
ap0nia, line 80:
```suggestion
    prerequisites: course.prerequisites as CoursePreview[],
    dependencies: course.dependencies as CoursePreview[],
```

apps/api/src/routes/v1/graphql/schema/instructors.graphql
```diff
@@ -1,3 +1,13 @@
+"An object that contains a subset of an instructor's metadata, for preview purposes."
+type InstructorPreview {
+  "The instructor's UCINetID."
+  ucinetid: String!
+  "The full name of the instructor."
+  name: String!
+  "The shortened name (or WebSoc name; e.g. ``SHINDLER, M.``) of the instructor."
+  shortenedName: String!
```
ap0nia, line 8: Maybe "displayName" or "shortName"? A past participle seems a bit verbose to me.

libs/db/prisma/schema.prisma
```diff
@@ -70,33 +70,37 @@ model Course {
   maxUnits          Float
   description       String      @db.Text
   departmentName    String
-  instructorHistory Json
+  instructorHistory String[]
```
ap0nia, line 73: What if there was a "History" table that joined "Course" and "Instructor"? It would have a primary key of (course ID, instructor ID, and term) or something.
Idk if it's worth the extra work, but it would enable the ability to query the history of an instructor, i.e. to see if they have a good track record.

libs/db/prisma/schema.prisma
```diff
@@ -70,33 +70,37 @@ model Course {
...
+  dependencies      Json        @default("[]")
   repeatability     String
   gradingOption     String
   concurrent        String
   sameAs            String
   restriction       String      @db.Text
   overlap           String
   corequisites      String
-  geList            Json
+  geList            String[]
```
ap0nia, line 88: One cool idea I've seen before is having a table of constants or enum of some sort. Then represent the GE List as a binary number. So if there are 8 GEs, the binary `10110000` would indicate that GEs 1, 3, 4 are represented by this entry.

libs/db/prisma/schema.prisma
```diff
@@ -70,33 +70,37 @@ model Course {
...
-  instructorHistory Json
+  instructorHistory String[]
+  instructors       Json        @default("[]")
   prerequisiteTree  Json
-  prerequisiteList  Json
+  prerequisiteList  String[]
   prerequisiteText  String      @db.Text
-  prerequisiteFor   Json
+  prerequisiteFor   String[]
+  prerequisites     Json        @default("[]")
```
ap0nia, line 79: I feel like this could actually be derived from a table? For example, a "Prerequisites" table that mapped course ID (prerequisite) to course ID (prerequisite for), uniquely keyed on those two columns.

To find a course's prerequisites given its course ID, search by the "prerequisite for" column with the same ID.
To find the courses that a course satisfies prerequisites for, search by the "prerequisite" column with the same ID.

I do realize that this is scrapped data and representing it as JSON is pretty minimal overhead on the backend complexity. But since we're scrapping it anyways, maybe we could store it optimally?

tools/registrar-scraper/src/instructor-scraper/index.ts
```diff
@@ -26,7 +26,7 @@ type InstructorsData = {
 };
 
 type InstructorsInfo = {
-  [ucinetid: string]: Instructor;
+  [ucinetid: string]: Omit<Instructor, "courses">;
```
ap0nia, line 29: I prefer using records if there's no information to preserve.
```suggestion
  Record<string, Omit<Instructor, "courses">>
```
 The main reason I use the mapped is to preserve interface information.

```ts
type MyType = {
  /** JSDoc for property a */
  a: string,
  /** JSDoc for property b*/
  b: string,
}

type DerivedType = Record<keyof MyType, any>
// Lost all of the JSDoc info.
type DerivedTypeA = DerivedType["a"]

type BetterDerived = { [k in keyof MyType]: MyType[k]}
// The JSDoc is preserved.
type BetterDerivedA = BetterDerived["a"]

If the key itself is a string, then there isn't any JSDoc information to preserve anyways.
```

tools/registrar-scraper/src/instructor-scraper/index.ts
```diff
@@ -139,7 +139,7 @@ export async function getInstructors(
       }
     });
   });
-  const instructorPromises: Promise<[string, Instructor]>[] = [];
+  const instructorPromises: Promise<[string, Omit<Instructor, "courses">]>[] = [];
```
ap0nia, line 142: Could we extract the `Omit<Instructor, "courses">` type? Maybe like "ScrappedInstructor"?

tools/registrar-scraper/src/index.ts
```diff
@@ -52,6 +52,67 @@ async function main() {
...
+        .map((x) => courses[x.replace(/ /g, "")])
+        .filter((x) => x)
+        .map(({ id, title, department, courseNumber }) => ({
+          id,
+          title,
+          department,
+          courseNumber,
+        })),
+    });
+  }
```
ap0nia, line 87: Functional programming 😎 
```suggestion
  const newCourses = Object.values(courses).map(course => {
    return {
      ...course,
      prerequisiteTree: course.prerequisiteTree as object,
      instructors: course.instructorHistory
        .map((x) => instructors[x])
        .filter((x) => x)
        .map(({ ucinetid, name, shortenedName }) => ({ ucinetid, name, shortenedName })),
      prerequisites: course.prerequisiteList
        .map((x) => courses[x.replace(/ /g, "")])
        .filter((x) => x)
        .map(({ id, title, department, courseNumber }) => ({
          id,
          title,
          department,
          courseNumber,
        })),
      dependencies: course.prerequisiteFor
        .map((x) => courses[x.replace(/ /g, "")])
        .filter((x) => x)
        .map(({ id, title, department, courseNumber }) => ({
          id,
          title,
          department,
          courseNumber,
        })),
    }
  })
```

tools/registrar-scraper/src/index.ts
```diff
@@ -52,6 +52,67 @@ async function main() {
...
```
ap0nia, line 103:
```suggestion
  const newInstructors = Object.values(instructors).map(instructor => {
    return {
      ...instructor,
      courseHistory: instructor.courseHistory as object,
      courses: Object.keys(instructor.courseHistory!)
        .map((x) => courses[x.replace(/ /g, "")])
        .filter((x) => x)
        .map(({ id, title, department, courseNumber }) => ({
          id,
          title,
          department,
          courseNumber,
        })),
    }
  })
```

ap0nia (changes requested):
Good stuff 💪 
I left mostly database design considerations lol. The code looks fine; once it's clear what how we'd like the database to be organized, I think it's good to merge.

libs/db/prisma/schema.prisma
ecxyzzy: Good idea, I think eventually we will want to migrate to this model.

ecxyzzy (commented):
Thanks for the feedback, I've incorporated most of the suggestions.

For posterity, we discussed the schema issues in DMs. We may consider in the future to use actual SQL relations, but the current solution provides the best performance since joins result in additional overhead.

apps/api/src/lib/utils.ts
```diff
@@ -1,5 +1,12 @@
 import { Course as PrismaCourse } from "@libs/db";
-import { Course, CourseLevel, GECategory, PrerequisiteTree } from "@peterportal-api/types";
+import {
+  Course,
+  CourseLevel,
+  CoursePreview,
+  GECategory,
+  InstructorPreview,
+  PrerequisiteTree,
+} from "@peterportal-api/types";
```
ap0nia, line 9: For clarity, I like to prefix any type-only imports with `type`. This is actually enforceable through tsconfig via the [importsNotUsedAsValues](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues) setting, so it's not too important to do right now. It would be better to turn on the setting and open a PR that fixes the errors.

tools/registrar-scraper/src/index.ts
```diff
@@ -52,6 +52,61 @@ async function main() {
...
+  const courses = Object.fromEntries((await prisma.course.findMany()).map((x) => [x.id, x]));
+  const instructors = Object.fromEntries(
+    (await prisma.instructor.findMany()).map((x) => [x.ucinetid, x]),
+  );
+  const newCourses = Object.values(courses).map((course) => ({
+    ...course,
+    prerequisiteTree: course.prerequisiteTree as object,
+    instructors: course.instructorHistory
+      .map((x) => instructors[x])
+      .filter((x) => x)
```
ap0nia, line 64: Slightly more efficient because we don't have to allocate the filter function multiple times, and a bit more legible because it's clear we're filtering for non-falsy values.

```suggestion
      .filter(notNull)
```

ap0nia (approved):
🚀

apps/api/src/lib/utils.ts
ecxyzzy: I like using `import type` as well but sometimes I miss it. Nice catch and thanks for letting me know that this exists; I'll make an issue for it.

ap0nia: Oh, yeah I guess it's supposed to be [verbatimModuleSyntax](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax) now.

Issue #98, opened 2023-09-09
Add DynamoDB-backed cache for select endpoints


ecxyzzy:
`@libs/ddb` be like:
![image](https://github.com/icssc/peterportal-api-next/assets/89349085/5ffdfc45-3006-4ec7-b04f-83927eb118fd)

---

Biggest offenders right now are probably grades and WebSoc, and those will definitely be cached. Feel free to suggest any other endpoints that might need caching below.

Blocked by #96.

ecxyzzy:
After further testing it appears that adding the cache generates a negligible performance boost. Maybe we just got good in five months, so going to put this in the icebox for now.

Issue #89, opened 2023-08-31
Clean up workspace


ecxyzzy:
- Use `workspace:^` for internal packages (see [here](https://github.com/icssc/peterportal-api-next/pull/68/files#r1310938105))
- Set up and fix workspace root per https://github.com/Thinkmill/manypkg
  - Remove `NODE_ENV=production` shim from Actions workflows

PR #126, opened 2024-02-02
feat: ✨ rework enrollment history endpoint

### LINKED ISSUE #121: Enrollment history resets on deploy
Not really sure what's happening here but this needs to be fixed before we deploy any new features.

ecxyzzy:
## Summary

Rework the schema related to the enrollment history endpoint, so newly scraped data is always inserted and never removed from the database. This will prevent enrollment history from being reset on deploy.

Also added some constraints regarding what parameters are required to look up enrollment history data that are consistent with ZotTracker's current restrictions.

## Next Steps

This endpoint will definitely benefit from a key-value cache per #98 so I need to get that done…

## Issues

Closes #121.


.github/workflows/deploy-prod.yml
```diff
@@ -80,27 +80,27 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+
+      - name: Setup Node.js and pnpm
+        uses: ./.github/actions/setup-node-and-pnpm
+
+      - name: Install dependencies
+        # Ensure NODE_ENV != production so pnpm will install devDependencies!!!
+        run: NODE_ENV=development pnpm install --frozen-lockfile
```
ap0nia, line 100: I was thinking of opening a PR with this new strat 💯 

```suggestion
        run: pnpm install --frozen-lockfile --prod false
```

https://pnpm.io/cli/install#--prod--p

apps/api/src/routes/v1/rest/enrollmentHistory/lib.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const transformResults = (
+  results: Array<WebsocEnrollmentHistory & { entries: WebsocEnrollmentHistoryEntry[] }>,
+): EnrollmentHistory[] =>
+  results.map(({ entries, ...x }) => ({
+    ...x,
+    sectionCode: x.sectionCode.toString().padStart(5, "0"),
+    meetings: x.meetings as Meeting[],
+    dates: entries.map(
+      (y) => `${y.date.getFullYear()}-${y.date.getMonth() + 1}-${y.date.getDate()}`,
```
ap0nia, line 12: I think it would be helpful to dedicate a function to this, so that way encoding/decoding the dates is clearer. e.g.

```ts
// Creates stringified representation of a websoc date.
function formatWebsocDate(...) { ... }

// Hypothetically, we could undo the stringifying.
function decodeWebsocDate(...) { }
```

apps/api/src/routes/v1/rest/enrollmentHistory/schema.ts
```diff
@@ -1,23 +1,30 @@
...
+    sectionType: z
+      .enum(sectionTypes, { invalid_type_error: "Invalid sectionType provided" })
+      .optional(),
+  })
+  .refine(
+    (x) =>
+      (x.department && x.courseNumber) ||
+      (x.sectionCode && x.year && x.quarter) ||
+      (x.instructor && x.courseNumber && x.year && x.quarter),
+  );
```
ap0nia, line 28: A message might be helpful here (I forgot the format for adding them in a refine).

```suggestion
  .refine(
    (x) =>
      (x.department && x.courseNumber) ||
      (x.sectionCode && x.year && x.quarter) ||
      (x.instructor && x.courseNumber && x.year && x.quarter),
  , { message: "Needs either ..." });
```

ap0nia (commented):
LGTM :rocket:

My recommendations are mostly insubstantial, so you can use them as you see fit.

.github/workflows/deploy-prod.yml
ecxyzzy: Per #89, eventually we want to make all root-level devDeps full dependencies so we don't have to add shims at all.

ap0nia (approved):
👍

PR #129, opened 2024-02-06
chore: 🔧 clean up workspace, add utils [skip ci]

### LINKED ISSUE #89: Clean up workspace

### LINKED ISSUE #90: Utils package
- https://github.com/icssc/peterportal-api-next/pull/68/files#r1310943542
- Move `APILambdaClient` here as well after #95 gets merged

ecxyzzy:
## Summary

* Changed version specifiers for all internal packages to `workspace:^`.
* Moved all workspace root devDependencies to full dependencies and removed production deployment shim. Also added `ignore-workspace-root-check` to the `.npmrc` so `pnpm` won't complain when trying to install a dependency in the workspace root.
  * I was going to add [@manypkg/cli](https://github.com/Thinkmill/manypkg), but they change the version specifier for all packages (even internal) from `workspace:^` to the actual version, which is not ideal for our usecase.
* Moved `sleep` and `notNull` to a new package `@libs/utils`.
* Added [`@typescript-eslint/consistent-type-imports`](https://typescript-eslint.io/rules/consistent-type-imports/), which enforces the `import type` syntax for type imports, to the `eslintrc`, and relinted the repo.
* Added `.then()` to async entrypoints, and removed dummy then callbacks (e.g. `(() => [])`).
## Issues
Closes #89; closes #90.


.eslintrc.cjs
```diff
@@ -19,6 +19,7 @@ const config = {
   ],
   rules: {
     "no-unused-vars": "off",
+    "@typescript-eslint/consistent-type-imports": "error",
```
ap0nia, line 22: I prefer using the tsconfig option [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax)

This is the TS error I get when I forgot to specify the type modifier:
image

.npmrc
```diff
@@ -1 +1,2 @@
 engine-strict = true
+ignore-workspace-root-check = true
```
ap0nia, line 2: tbh, I'm ok with having the annoying warning message because installing packages in the workspace root should ideally be uncommon. I'm also used to doing `pnpm i -Dw <package>` whenever I need to do this specifically 🤷

libs/utils/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@libs/utils",
+  "version": "0.0.0",
```
ap0nia, line 3: Example description to add; either a description or a README might be helpful for contextualizing what the package should be used for and what functionality can be added in the future.

```suggestion
  "version": "0.0.0",
  "description": "a collection of miscellaneous utilities used by peterportal-api request handlers"
```

package.json
```diff
@@ -18,7 +18,7 @@
     "lint": "eslint --fix .",
     "test": "turbo run test"
   },
-  "devDependencies": {
+  "dependencies": {
```
ap0nia, line 21: Is there a benefit to doing this? I prefer using `devDependencies` because none of the packages are used in the built modules. Functionally, `devDependencies` and `dependencies` are identical except when specifically `NODE_ENV=production`, so I like to use them as semantic indicators.

ap0nia (approved):
I'm not too opinionated on devops atm, so I think it's fine to implement whatever changes you're interested in, and then open up a discussion later if that's desired. 👍

PR #132, opened 2024-02-17
feat: ✨ add calendar scraper and dump endpoint

### LINKED ISSUE #111: Create seeding tool/scraper for calendar endpoint
Right now, the calendar endpoint fetches data from the registrar on demand if it is not present in the database. This is less than ideal from both a response latency and separation-of-concerns standpoint. Furthermore, this data could simply be scraped once a year, and on an *ad hoc* basis during the year as inaccuracies are identified, similar to the registrar scraper.

### LINKED ISSUE #124: Calendar Dump for Term Start and Finals Start
As of now, the Calendar endpoint requires specification of quarter and year:
https://docs.api-next.peterportal.org/docs/developers-guide/rest-api/reference/calendar

(This is in contrast to the terms endpoint, which produces every available term name). It would be useful to have a similar option for Calendar such that the information available could be something like a map of each term to the corresponding information for finals/term dates. 

As a follow up issue, a programmatic way to access the date the schedule of classes is available would also be of use.

ecxyzzy:
## Summary
- Add automated scraper for the calendar endpoint, which will run every three months (probably on the generous side but you never know).
- Calling `/v1/rest/calendar` with no params will now dump all terms in the database. The GraphQL query `allTermDates` accomplishes the same.
- Add the field `socAvailable` to the response type, which indicates when the Schedule of Classes will become available for that term.

## Issues
Closes #111; closes #124.

### DIFF
--- a/apps/api/src/routes/v1/graphql/resolvers.ts
+++ b/apps/api/src/routes/v1/graphql/resolvers.ts
@@ -4,6 +4,7 @@ import { geTransform, proxyRestApi } from "./lib";
 
 export const resolvers: ApolloServerOptions<BaseContext>["resolvers"] = {
   Query: {
+    allTermDates: proxyRestApi("/v1/rest/calendar"),
     calendar: proxyRestApi("/v1/rest/calendar"),
     course: proxyRestApi("/v1/rest/courses", { pathArg: "courseId" }),
     courses: proxyRestApi("/v1/rest/courses", { argsTransform: geTransform }),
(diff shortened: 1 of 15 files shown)


adcockdalton (changes requested):
I cannot discern where the issue might be in the code as I have little to no experience with webscraping, but the output for the finals start and end fields is identical.

I'm not sure where that issue could be sourced, but the output is wrong there. Otherwise looks great, thank you so much!

ecxyzzy:
Yeah I narrowed down the issue, it's just a data integrity problem that I need to resolve as I may have imported faulty data into the dev database. I'll do that at some point and re-request your review then.

Issue #115, opened 2023-12-04
Convert REST response type to tagged union


ecxyzzy:
Currently, to determine whether a `RawResponse<T>` is a `Response<T>` or an `ErrorResponse`, we test whether the `payload` field exists. This is kind of cumbersome and is why we ship an `isErrorResponse` function with the types package, but we could just add `success: true` to the former variant of `RawResponse<T>` and `success: false` to the latter.

This would make error checking as simple as testing `res.success`, and would also allow consumers of the types package to only add it as a `devDependency` (not sure if the latter matters as much though).

PR #134, opened 2024-02-29
feat: ✨ changed error response to success flag

jordany33:
re #115


ecxyzzy (changes requested):
Sorry for my late response.

Just adding `success: boolean` to `BaseResponse` doesn't actually work, as the checks show, because `RawResponse<T>` needs to be a [tagged union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions). Should be a relatively straightforward fix, but let me know if you have any issues.

ecxyzzy (approved):
LGTM, thanks for working on this!

PR #135, opened 2024-02-29
feat: ✨ refine course history scraper

### LINKED ISSUE #113: Scrape course history from WebReg EnrollHist
In the registrar scraper, the set of terms in which a course is offered is determined by taking all instructors that taught that course, and determining in which term(s) the instructors taught the course. This works for some courses, but it seems that instructors listed on the General Catalogue do not teach certain courses, and as such those courses do not have the correct set of terms. See https://github.com/icssc/peterportal-client/pull/351#discussion_r1406908650.

Instead, we could consider using [EnrollHist](https://www.reg.uci.edu/perl/EnrollHist.pl) in addition to or in place of what we do currently.

cokwong:
## Summary
- Term history for courses were originally cross-referenced from scraped instructors' data. Some courses have instructors not in the directory or just `Staff` resulting in missing terms.

- Implemented crawler to scrape course history of a department from [EnrollHist](https://www.reg.uci.edu/perl/EnrollHist.pl).

- The new data is joined with the original results from instructors' course history (in case crawler fails).

## Issues
- Closes #113


ecxyzzy (approved):
LGTM, thanks for working on this!

The instructors' course history isn't updated but to be fair the issue also didn't mention it. I'll create a new issue to track that.

PR #141, opened 2024-03-11
feat: fuzzy SaaS proof of concept (WIP)

### LINKED ISSUE #131: RFC: Fuzzy SaaS and Rewrite
## Current State of Fuzzy Search

It works and seems reasonably fast somehow, but is a bit of a cobbled together mess. This is of course not entirely surprising considering it _was_ hacked together within the span of a week. There are some weird edge cases that have come up because of how it actually works under the hood, and at this point I'm not entirely sure at first glance how to address them since the code is so poorly documented (thanks to me from two years ago).

## Question

Fuzzy search will definitely be reworked—that is not really up for debate. The main decision to be made is whether to continue offering it as an `npm` package, or creating an API endpoint that encapsulates the functionality.

I'm also open to any additional solutions that are none of the above or incorporate elements of both.

### Pros
* Ease of maintenance: no more package updates required
* Smaller bundle size for consumers
* Lower entry barrier, since searching would just be another API call

### Cons
* Increased overhead/latency from making HTTP requests
* Difficulty of implementation

## Stakeholders

@EricPedley @ap0nia @js0mmer

ecxyzzy:
## Summary

Implement fuzzy search as a service through REST and GraphQL (`searchCourses` and `searchInstructors`, since GraphQL doesn't support tagged union/sum types).

The total number of results is returned in the `count` field in the payload, and the result set (containing at most the requested number of results starting from the offset) is returned in the `results` field.

### Parameters
- `query`: The string to fuzzy-search. Required.
- `resultType`: The type of result to return exclusively, `course` or `instructor`. Both types of results will be returned if left undefined.
- `limit` and `offset`: For pagination purposes, defaults to 10 and 0 respectively.

## Issues

Closes #131.

### DIFF
--- a/apps/api/bronya.config.ts
+++ b/apps/api/bronya.config.ts
@@ -121,6 +121,7 @@ export const esbuildOptions: BuildOptions = {
           path: args.path,
           namespace,
         }));
+        build.onResolve({ filter: /virtual:search/ }, (args) => ({ path: args.path, namespace }));
         build.onLoad({ filter: /virtual:courses/, namespace }, async () => ({
           contents: `export const courses = ${JSON.stringify(
             Object.fromEntries(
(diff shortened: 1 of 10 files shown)


js0mmer (commented):
just from some quick tests, performance looks good but some results are unexpected. searching for "ics" seems to be a problem since the course names are actually I&C SCI, although I've managed to have a similar issue with art. I think some kind of weighting so the department name is more important than the course title could help with the results here.

see https://github.com/icssc/peterportal-client/pull/466#issuecomment-2067450694 and https://github.com/icssc/peterportal-client/pull/466#issuecomment-2067451941

PR #147, opened 2024-05-19
refactor(api): ♻️ degrees endpoint

ap0nia:
Just some adjustments I thought would be neater. Too much to include in the GitHub comments.


ecxyzzy (approved):
Endpoint changes lgtm, definitely cleaner than whatever it was that I implemented lol. Also epic Zod schema transform to tagged union tech :rocket:

